PR #20, opened 2023-03-04
Reject vote with reason on shutter proposals

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot-hub/issues/491
- Should not accept `reason` in vote if shutter is enabled for proposal
- Same code as in https://github.com/snapshot-labs/snapshot-hub/pull/498
- Also changes default port in .env.example file


src/writer/vote.ts
```diff
@@ -37,6 +37,7 @@ export async function verify(body): Promise<any> {
     return Promise.reject('not in voting window');
 
   if (proposal.privacy === 'shutter') {
+    if (msg.payload.reason?.length) return Promise.reject('reason not allowed with shutter');
```
ChaituVR, line 40:
```suggestion
    if (msg.payload.reason) return Promise.reject('reason not allowed with shutter');
```

bonustrack (approved):
utACK

samuveth (approved):
utACK

PR #21, opened 2023-03-04
 Allow ENS controller to be snapshot space controller

ChaituVR:
Part of https://github.com/snapshot-labs/snapshot/issues/3472
---------------

**!IMPORTANT:**  So one thing to note is if the `snapshot` text record is IPFS or URL, we still allow new settings from the ENS controller, I think it should be fine

---------------
TODO:
- [x] Test it with mainnet and goerli ENS names
- [x] with different ENS controllers and snapshot existing controllers
- [x] Test with IPFS in snapshot text record


bonustrack (approved):
utACK

PR #23, opened 2023-03-06
Calculate votes count and scores even if shutter is enabled

ChaituVR:
Works only with https://github.com/snapshot-labs/snapshot.js/pull/791

### DIFF
--- a/src/scores.ts
+++ b/src/scores.ts
@@ -135,8 +135,10 @@ export async function updateProposalAndVotes(proposalId: string, force = false) 
   if (proposal.scores_state === 'final') return true;
 
   if (!force && proposal.privacy === 'shutter') {
-    if (proposal.state === 'closed') await getDecryptionKey(proposal.id);
-    return true;
+    if (proposal.state === 'closed') {
+      await getDecryptionKey(proposal.id);
+      return true;
+    }
   }
 
   // Get votes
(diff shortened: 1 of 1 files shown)


src/scores.ts
```diff
@@ -166,7 +166,18 @@ export async function updateProposalAndVotes(proposalId: string, force = false)
...
 
   // Get results
-  const voting = new snapshot.utils.voting[proposal.type](proposal, votes, proposal.strategies);
+  const encryptedChoice = proposal.privacy === 'shutter' && proposal.state !== 'closed';
+
+  const voting = new snapshot.utils.voting[proposal.type](
+    proposal,
+    votes,
+    proposal.strategies,
+    undefined,
```
samuveth, line 175: We can remove `undefined` and `{ encryptedChoice }` as they are no longer needed with https://github.com/snapshot-labs/snapshot.js/pull/791

src/scores.ts
```diff
@@ -166,7 +166,18 @@ export async function updateProposalAndVotes(proposalId: string, force = false)
...
+    {
+      encryptedChoice
+    }
+  );
+
```
samuveth, line 180:
```suggestion

  const voting = new snapshot.utils.voting[proposal.type](
    proposal,
    votes,
    proposal.strategies
  );

```

src/scores.ts
```diff
@@ -167,6 +167,7 @@ export async function updateProposalAndVotes(proposalId: string, force = false)
 
   // Get results
   const voting = new snapshot.utils.voting[proposal.type](proposal, votes, proposal.strategies);
+
```
ChaituVR, line 170:
```suggestion
```

samuveth (approved):
utACK

PR #34, opened 2023-03-20
Delete space action

ChaituVR:
Closes https://github.com/snapshot-labs/snapshot-hub/issues/563

**Important**
Admin task:
- [x]  Run this command:
```SQL
ALTER TABLE
  `spaces`
ADD
  COLUMN `deleted` INT NOT NULL DEFAULT 0 After verified;
```

### DIFF
--- a/src/helpers/actions.ts
+++ b/src/helpers/actions.ts
@@ -32,11 +32,13 @@ export async function getProposal(space, id) {
   return proposal;
 }
 
-export async function getSpace(id) {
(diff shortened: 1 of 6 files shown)


bonustrack:
@ChaituVR Added the column `deleted`. Code looks good, we may want to handle also the case where someone create a space that was deleted previously.

bonustrack (approved):
utACK, added the index

PR #47, opened 2023-05-11
feat(ipfs body): add EIP712 hash to ipfs body

### LINKED ISSUE #45: feat: add hash in the payload of pinned message
Currently we store message payload on IPFS with this format: 
https://snapshot.mypinata.cloud/ipfs/bafkreiduyg7mxfcizibhihejr7s7e6e7ow4ikp7ctom36zlwncrfi4a6ra
We should add the message `hash` after `sig`, this will make it easier to connect an IPFS file to the message id (as the id is actually the message hash). We should also test it on Signatorio site to make sure we don't break the verification.

Todmy:
Closes #45

### DIFF
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -141,8 +141,9 @@ export default async function ingestor(body) {
   let pinned;
   let receipt;
   try {
-    const { sig, ...restBody } = body;
+    const { addrress, sig, ...restBody } = body;
     const ipfsBody = {
+      addrress,
       sig,
       hash: id,
       ...restBody
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -141,7 +141,12 @@ export default async function ingestor(body) {
   let pinned;
   let receipt;
   try {
-    const ipfsBody = { ...body, hash: id };
+    const { sig, ...restBody } = body;
+    const ipfsBody = {
+      sig,
+      hash: id,
+      ...restBody
+    };
     [pinned, receipt] = await Promise.all([pin(ipfsBody), issueReceipt(body.sig)]);
   } catch (e) {
     return Promise.reject('pinning failed');
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -141,7 +141,8 @@ export default async function ingestor(body) {
   let pinned;
   let receipt;
   try {
-    [pinned, receipt] = await Promise.all([pin(body), issueReceipt(body.sig)]);
+    const ipfsBody = { ...body, hash: id };
+    [pinned, receipt] = await Promise.all([pin(ipfsBody), issueReceipt(body.sig)]);
   } catch (e) {
     return Promise.reject('pinning failed');
   }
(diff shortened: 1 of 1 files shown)


bonustrack:
The param `hash` should appear just after `sig`, not in the end of the payload. Also can you confirm it will work with Signatorio website? You can find a Signatorio link on Snapshot this way:
image

src/ingestor.ts
```diff
@@ -141,7 +141,14 @@ export default async function ingestor(body) {
...
   try {
-    [pinned, receipt] = await Promise.all([pin(body), issueReceipt(body.sig)]);
+    const { addrress, sig, ...restBody } = body;
+    const ipfsBody = {
+      addrress,
+      sig,
+      hash: id,
+      ...restBody
+    };
+    [pinned, receipt] = await Promise.all([pin(ipfsBody), issueReceipt(body.sig)]);
```
ChaituVR, line 151:
```suggestion
    const { sig } = body;
    const ipfsBody = {
      hash: id,
      ...restBody
    };
    [pinned, receipt] = await Promise.all([pin(ipfsBody), issueReceipt(body.sig)]);
```

Also, I don't think we can verify this signature without removing the `hash` from params, as mentioned in 
https://github.com/snapshot-labs/snapshot-sequencer/pull/47#issuecomment-1544561913 you tried?

Todmy:
this is how it looks now:

Screenshot 2023-05-27 at 10 52 25
Screenshot 2023-05-27 at 10 53 54

Is it ok? Do you want to change the way how `signatorio` displays the message?

bonustrack:
Did you tested with such URL (with the IPFS cid) `https://signator.io/view?ipfs=bafkreigklpan44he4sqkkt23sq3vah63acqibianifcslsh4cczsziziq4` ? It should work that way, this is the link we use on snapshot.org

Todmy:
> Did you tested with such URL (with the IPFS cid) `https://signator.io/view?ipfs=bafkreigklpan44he4sqkkt23sq3vah63acqibianifcslsh4cczsziziq4` ? It should work that way, this is the link we use on snapshot.org

yes, I have used this one. The same you showed in the first comment 

237793082-4cd84483-8356-4359-ab3b-068277f449b4

This is the link to the signator from the newly created proposal ([link](https://signator.io/view?ipfs=bafkreielu23nbdsxiyejsk7u2tfnecj56p47jmblch7v4s2lys4dmtb43a)).

Please let me know if I miss something

src/ingestor.ts
```diff
@@ -141,7 +141,14 @@ export default async function ingestor(body) {
   let pinned;
   let receipt;
   try {
-    [pinned, receipt] = await Promise.all([pin(body), issueReceipt(body.sig)]);
+    const { addrress, sig, ...restBody } = body;
+    const ipfsBody = {
+      addrress,
```
bonustrack, line 146: There is a type with `addrress` should be `address`

bonustrack (approved):
utACK

PR #50, opened 2023-05-21
feat(ci/cd): add docker to the project

Todmy:
This PR add files that allow to run project in docker container


.env.example
bonustrack, line 1: These values are included in DATABASE_URL already, can't we reuse them instead of creating new env var?

Todmy: I did this because I configured the container in a way it automatically creates schemas in DB. This allows us to define schemas in a very flexible way and run migrations on DB. 

Another way to do it is to change `schema.sql` itself, but it will lock the name of DB and it will be hard to change it. 

The third way is to create multiple scripts that will automatically parse everything and pick the name of DB from `DATABASE_URL` - I don't like this approach because it requires more coding and magic, so it will be more fragile.

PR #54, opened 2023-05-25
Ignore score calculation if more than 100k votes

ChaituVR:


src/scores.ts
```diff
@@ -143,6 +143,13 @@ export async function updateProposalAndVotes(proposalId: string, force = false)
     proposalId === '0xda4f201a37ea08cf1892418e7b9e88f5687a68dbdc96c3ab22abaa1c7244648e';
   if (isLinea) return;
 
+  // Ignore score calculation if proposal have more than 100k votes and scores_updated greater than 1 minute
+  const ts = Number((Date.now() / 1e3).toFixed());
+  if (proposal.votes > 0 && proposal.scores_updated > ts - 60) {
```
aurelianoB, line 148: based on the comment, I think you want to limit if votes are greater than 100k, not 0
```suggestion
  if (proposal.votes > 100000 && proposal.scores_updated > ts - 60) {
```

bonustrack (approved):
utACK

PR #69, opened 2023-06-13
Check schema for statement

ChaituVR:
Merge only after https://github.com/snapshot-labs/snapshot.js/pull/850


bonustrack (approved):
utACK

PR #71, opened 2023-06-14
feat(proposal start): restrict start time of proposal to be greater than now

Todmy:
[#334](https://github.com/snapshot-labs/snapshot-hub/issues/334)

### DIFF
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -166,10 +166,6 @@ export default async function ingestor(req) {
   try {
     const { address, sig, ...restBody } = JSON.parse(JSON.stringify(body));
 
-    if (restBody.data?.message?.start) {
-      restBody.data.message.start = Math.max(restBody.data.message.start, ts);
-    }
-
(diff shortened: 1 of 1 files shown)


bonustrack:
Did you tried cast a vote on UI with this change?

src/ingestor.ts
```diff
@@ -96,13 +96,13 @@ export default async function ingestor(req) {
...
 
-  if (type === 'proposal')
+  if (type === 'proposal') {
     payload = {
       name: message.title,
       body: message.body,
       discussion: message.discussion || '',
       choices: message.choices,
-      start: message.start,
+      start: message.start > ts ? message.start : ts,
```
ChaituVR, line 105: @Todmy I think It should throw an error instead of modifying the `start` users may complain comparing with IPFS file

Todmy:
after a discussion with @ChaituVR we got the consensus that we shouldn't change `start` of a proposal on the backend since it could lead to bugs.

PR #79, opened 2023-07-05
feat: add custom limit for flagged and unverified spaces

bonustrack:
Adding new limit for daily and monthly proposal created for flagged and unverified spaces.


ChaituVR (approved):
utAck

PR #80, opened 2023-07-07
chore: add tests

wa0x6e:
- [x] Add tests list of test to implement
- [x] Setup CI
- [x] Add lint and test to CI

The list of test to implement has been added.
This PR can be reviewed and pulled any time, the rest of the tests will be implemented little by little over time, in subsequent PR

## Considerations

Tests only run on local for now, due to snapshot.js SDK timing out all jest tests


package.json
```diff
@@ -3,9 +3,12 @@
   "version": "0.1.3",
   "license": "MIT",
   "scripts": {
-    "lint": "eslint . --ext .ts --fix",
+    "lint": "yarn lint:nofix --fix",
+    "lint:nofix": "eslint . --ext .ts",
```
Todmy, line 7: nit
```suggestion
    "lint:fix": "yarn lint --fix",
    "lint": "eslint . --ext .ts",
```

package.json
```diff
@@ -3,9 +3,12 @@
...
     "dev": "nodemon src/index.ts",
-    "start": "ts-node src/index.ts"
+    "start": "ts-node src/index.ts",
```
Todmy, line 9: can we add here `--transpile-only`?

test/schema.sql
```diff
@@ -0,0 +1,144 @@
...
+  space VARCHAR(64) NOT NULL,
+  about TEXT,
+  statement TEXT,
+  created INT(11) NOT NULL,
+  updated INT(11) NOT NULL,
+  PRIMARY KEY (delegate, space),
+  INDEX ipfs (ipfs),
+  INDEX created (created),
+  INDEX updated (updated)
+);
```
Todmy, line 144: nit
```suggestion
);

```

package.json
wa0x6e: I also wished that could happen :P

This was kept like that because most of the project was using `yarn lint` with --fix by default. This is a technical debt, a similar PR with this fix was rejected in the past.

Maybe a future organization wide PR will fix that, on the DX side

package.json
wa0x6e: This is a tests-focused pr. You could open another PR for that

test/schema.sql
ChaituVR, line 1: Tests should reuse the original schema, not ideal to maintain this in two places

ChaituVR: Oh is it because schema is on snapshot-hub? 🙈 got it

Todmy (approved):
utACK

PR #84, opened 2023-07-12
Use proposal strategies if basic validation doesn't have strategies

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/pull/4053

Right now if we are not using custom strategies in basic validation, it fails, we should fallback to proposal strategies instead

How to test:
- Change your space settings to use basic validation and don't use any custom strategies
- Try to create a proposal and vote on it
- It will fail earlier, it should pass through with these changes


Todmy (approved):
utACK

PR #86, opened 2023-07-15
chore: add sentry

wa0x6e:
- Setup project to send errors to Sentry
- Send all caught errors to sentry (basically, call `capture(e)` each time there is a `catch (e)`


Todmy (approved):
utACK

PR #93, opened 2023-07-27
chore: add sentry to CI

wa0x6e:
- Create sourcemaps on build
- Upload the sourcemaps and create a Sentry release when deploying on main/master branch


Todmy (approved):
utACK

PR #94, opened 2023-07-27
chore: run prod app with node instead of ts-node

wa0x6e:

The current production app is run directly on typescript file with `ts-node`, instead of `node` on compiled js files.

This result in up to 10x more memory consumption.

## 💊 Fixes / Solution

- Run app on compiled js files

## 🚧 Changes

- Build js files
- Run app on js files

## 🛠️ Tests

- Run `yarn build`
- Run `yarn start`
- It should start the app


ChaituVR (approved):
utAck

PR #97, opened 2023-07-28
fix: allow space members to skip active proposal limit

aurelianoB:
Add exception for members of a space (author, moderator or admin) to active proposal limit


ChaituVR (approved):
utAck

PR #98, opened 2023-07-29
fix: decrease active proposal limit for non-members

aurelianoB:
Reducing limit back after adding exception for space members https://github.com/snapshot-labs/snapshot-sequencer/pull/97


ChaituVR (approved):
UtAck

PR #100, opened 2023-08-01
hotfix: temporarily increase ecosystem daily proposal limit

aurelianoB:
Temporary fix to allow a slightly higher daily proposal limit for ecosystem spaces. To be reverted soon.


ChaituVR (approved):
utAck

PR #111, opened 2023-08-07
feat(env): add ability to configure server through .env

Todmy:


src/scores.ts
```diff
@@ -43,13 +43,14 @@ async function getVotes(proposalId: string): Promise<any[] | undefined> {
  * This should be implemented in snapshot.js, leading to either a breaking change or a new
  * function, e.g. named getFullScores while getScores still returns just the scores prop.
  */
+const scoreApiEndpoint = process.env.SCORE_API_URL || 'https://score.snapshot.org';
```
wa0x6e, line 46: Shouldn't this be named `scoreApiURL`, like the rest ?

wa0x6e (changes requested):

src/scores.ts
Todmy: fixed

wa0x6e (approved):
utAck

PR #112, opened 2023-08-09
Use getScores from snapshot.js

ChaituVR:
Depend on https://github.com/snapshot-labs/snapshot.js/pull/882


wa0x6e (commented):
I believe this also requires an update of the snapshot.js dependency to be complete ?

wa0x6e (approved):
utAck

PR #127, opened 2023-08-24
feat: Store user space activty

wa0x6e:
From https://www.notion.so/snapshotlabs/S-delegation-part-2-2b3186d8a33b4fd8b90c6575b6bf8000#bc7c1c9c785c46bdad69b4c8337771ca and https://github.com/snapshot-labs/snapshot-hub/pull/653

- [x] Update the `votes_count` on new votes
- [x] Update the `proposals_count` on new proposal
- [x] Update the `proposals_count` on proposal deletion
- [x] Update the `votes_count` on proposal deletion
- [x] Update the `proposals_count` on space soft-deletion
- [x] Update the `votes_count` on space soft-deletion
- [x] Add a script to generate all counters

## Use-cases not handled, due to them not being done through sequencer

- ~~Update the `votes_count` on space re-activation~~ (space should be hard-deleted)
- ~~Update the `proposals_count` on space re-activation~~ (space should be hard-deleted)
- ~~Update the `votes_count` on space migration~~ (space migration should be deprecated)
-  ~~Update the `proposals_count` on space migration~~ (space migration should be deprecated)


src/writer/proposal.ts
```diff
@@ -223,5 +223,16 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
...
+  if (result.affectedRows > 0) {
+    await db.queryAsync(
+      `
+      INSERT INTO user_space (space_id, user_id, proposals_count)
+      VALUES(?, ?, 1)
+      ON DUPLICATE KEY UPDATE proposals_count = proposals_count + 1
+      `,
+      [space, author]
+    );
+  }
```
ChaituVR, line 237: Could we merge all queries into one query, I worry about DB network errors after executing the first one and failing on the second one. then we will have wrong counts

wa0x6e: How would you merge it?

the second one depends on the result of the first one, as we should confirm there’s really a new row, else we skip the counter update.

the lack of transaction and counters un syncing has been raised, and @less has acknowledged that it’s an inconvenient we can leave with

wa0x6e: Failure, or duplicate

ChaituVR: But even if we merge, if the first query fails, the second one also fails no?

wa0x6e: Yes, unless it throws an exception. Second query executes only if there is one successful insert

src/writer/vote.ts
```diff
@@ -176,7 +176,11 @@ export async function action(body, ipfs, receipt, id, context): Promise<void> {
     );
   } else {
     // Store vote in dedicated table
-    await db.queryAsync('INSERT IGNORE INTO votes SET ?', params);
+    const result = await db.queryAsync('INSERT IGNORE INTO votes SET ?', params);
+
+    if (result.affectedRows > 0) {
+      await incrementVotesCount(msg.space, voter);
```
ChaituVR, line 182: What if this fails due to some network issue after inserting?

wa0x6e: If first insert query succeed, then something happens, and vote increment fail, then the counter will not increment.

As we're not using query transaction, we're not protected against this use case.

src/writer/delete-space.ts
```diff
@@ -21,6 +26,8 @@ export async function action(body): Promise<void> {
 
   try {
     await markSpaceAsDeleted(space);
+    await refreshProposalsCount(space);
+    await refreshVotesCount(space);
```
ChaituVR, line 30: These functions accept `string[]`? 🤔 Here space is string, right?

src/writer/vote.ts
wa0x6e: We can run cron to refresh some counters with the full computation (like in the generate scripts), instead of `+1`, to resync the counters

There's another way to keep the counters in sync, by using SQL triggers.

As for the solution used in keycard, it'll work only if we're 100% that the first query do insert in all cases, which does not seem the case, since it's using `INSERT IGNORE `, which imply that in some cases, it'll succeed, but not insert anything. In that case, we should not run the second query. Unless we're using a more complex SQL, with condition etc ...

scripts/generate_user_space_activities_counters.ts
```diff
@@ -0,0 +1,45 @@
...
+  const proposalsCountRes = await refreshProposalsCount();
+  console.log(proposalsCountRes);
+
+  console.log('Inserting/Updating votes_count');
+  let page = 0;
+  const batchSize = 1000;
+  const results: any[] = [];
+
+  while (true) {
+    const result = await db.queryAsync('SELECT id FROM spaces ORDER BY RAND () LIMIT ? OFFSET ?', [
```
ChaituVR, line 17: With `RAND` we may get the spaces that are already refreshed right?

src/helpers/actions.ts
```diff
@@ -52,3 +52,70 @@ export async function markSpaceAsDeleted(space: string) {
...
+  return db.queryAsync(
+    `
+      INSERT INTO user_space_activities (votes_count, user_id, space_id)
+        (SELECT * FROM (
+          SELECT COUNT(id) AS votes_count, voter, space
+          FROM votes
+          JOIN spaces ON spaces.id = votes.space
+          WHERE spaces.deleted = 0 AND space IN (?)
+          GROUP BY voter, space
+        ) AS t)
```
ChaituVR, line 82: Assume a space like stgdao with a lot of votes

scripts/generate_user_space_activities_counters.ts
wa0x6e: Is there a way to retrieve the spaces, sorted by votes_count ?

scripts/generate_user_space_activities_counters.ts
```diff
@@ -0,0 +1,45 @@
...
+
+(async () => {
+  try {
+    await main();
+    process.exit(0);
+  } catch (e) {
+    console.error(e);
+    process.exit(1);
+  }
+})();
```
ChaituVR, line 45: Do we want to run it every time the server starts? 🙈 We have 5 containers running. this may increase the load on the server

scripts/generate_user_space_activities_counters.ts
```diff
@@ -0,0 +1,45 @@
+import 'dotenv/config';
+import { refreshProposalsCount, refreshVotesCount } from '../src/helpers/actions';
+import db from '../src/helpers/mysql';
+
+// Usage: yarn ts-node scripts/generate_user_space_activities_counters.ts
+async function main() {
+  console.log('Inserting/Updating proposals_count');
+  const proposalsCountRes = await refreshProposalsCount();
```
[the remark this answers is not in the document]
wa0x6e: Fixed, added in batch

scripts/generate_user_space_activities_counters.ts
ChaituVR: There is no way to run manually on a single container I think 🤔 if we trigger it on DO's console will it continue to run even if we close browser?

src/writer/delete-space.ts
```diff
@@ -21,6 +26,8 @@ export async function action(body): Promise<void> {
 
   try {
     await markSpaceAsDeleted(space);
+    await refreshProposalsCount([space]);
+    await refreshVotesCount([space]);
```
ChaituVR, line 30: Now that we handle them separately. we should delete the count instead of refresh?

because we have `WHERE spaces.deleted = 0` condition 🤔

wa0x6e: That's correct, that will come with the PR which handle the space deletion/archive. As long as the soft-delete is here, this is still needed

scripts/generate_user_space_activities_counters.ts
wa0x6e: Did no try either. In worst case, we leave the browser open until the task end, or we pipe the command to null (add `> &` at the end, which send the task to the background).

Ideally, we should handle that with a queue system (was proposed in a pitch in cycle 3)

test/schema.sql
```diff
@@ -164,3 +164,14 @@ CREATE TABLE messages (
   INDEX type (type),
   INDEX receipt (receipt)
 );
+
+CREATE TABLE user_space_activities (
+  user_id varchar(64) NOT NULL,
+  space_id varchar(64) NOT NULL,
```
bonustrack, line 170: SQL command should be uppercase, also we should use same naming than what we have on others tables 
```suggestion
  user VARCHAR(64) NOT NULL,
  space VARCHAR(64) NOT NULL,
```

test/schema.sql
```diff
@@ -164,3 +164,14 @@ CREATE TABLE messages (
...
+  votes_count smallint unsigned NOT NULL DEFAULT '0',
+  proposals_count smallint unsigned NOT NULL DEFAULT '0',
+  UNIQUE KEY user_id_space_id (user_id,space_id),
```
bonustrack, line 173: Why not using a primary key?

src/helpers/actions.ts
```diff
@@ -52,3 +52,72 @@ export async function markSpaceAsDeleted(space: string) {
...
+  );
+}
+
+export function incrementProposalsCount(space_id: string, user_id: string) {
+  return db.queryAsync(
+    `
+      INSERT INTO user_space_activities (space_id, user_id, proposals_count)
+      VALUES(?, ?, 1)
+      ON DUPLICATE KEY UPDATE proposals_count = proposals_count + 1
+    `.replace(/\n/gm, ' '),
```
bonustrack, line 108: What is this for?

wa0x6e: The query is split on multiple line. This end up inserting a new line character `\n` on each new line, which make the SQL query invalid. The `replace()` will remove those invalid new lines characters.

Alternatively, we could also use simple string with simple/double quote, and use concatenation like

```js
'INSERT INTO user_space_activities (space_id, user_id, proposals_count)' +
'VALUES(?, ?, 1)' +
'ON DUPLICATE KEY UPDATE proposals_count = proposals_count + 1'
```

The concatenation solution make copy/paste more difficult, when trying to debug SQL query.

Or have the same query on the same line like this

```js
INSERT INTO user_space_activities (space_id, user_id, proposals_count VALUES(?, ?, 1) ON DUPLICATE KEY UPDATE proposals_count = proposals_count + 1
```

This reduce readability on complex queries.

Will be happy to migrate to any of the alternatives.

test/schema.sql
wa0x6e: fixed

test/schema.sql
wa0x6e: fixed

scripts/generate_user_space_activities_counters.ts
wa0x6e: edited to process one space at a time

src/helpers/actions.ts
wa0x6e: Removed

wa0x6e:
- Branch update with master
- Counter updates query has been merged together to be executed alongside the one inside the writers (since now, each insert is guaranteed to be a real insert, since the removal of `IGNORE`)

test/schema.sql
```diff
@@ -179,3 +179,14 @@ CREATE TABLE messages (
   INDEX type (type),
   INDEX receipt (receipt)
 );
+
+CREATE TABLE user_space_activities (
```
ChaituVR, line 183: Maybe `user_activity`

wa0x6e: Or user_stats ?

src/writer/delete-proposal.ts
```diff
@@ -20,11 +20,20 @@ export async function verify(body): Promise<any> {
...
-  `;
-  await db.queryAsync(query, [id, id]);
+  await db.queryAsync(
+    `
+    DELETE FROM proposals WHERE id = ? LIMIT 1;
+    DELETE FROM votes WHERE proposal = ?;
+    UPDATE user_space_activities
+      SET proposals_count = proposals_count - 1
+      WHERE user = ? AND space = ?
+      LIMIT 1;
```
ChaituVR, line 33: What happens to votes_count of user

wa0x6e: Computed via refreshVotes below

test/schema.sql
ChaituVR: or user_data or user_counts 😋  `user_activity` sounds good because we can add more activity here

wa0x6e: PRIMARY KEY is a combination of `user` and `space`, so this table can only track space related user's activities, and only counters.

ChaituVR: humm,, go for anything simple or just `activity` Last option is `user_space_activity`

scripts/generate_user_space_activities_counters.ts
```diff
@@ -0,0 +1,38 @@
...
+import { refreshProposalsCount, refreshVotesCount } from '../src/helpers/actions';
+import db from '../src/helpers/mysql';
+
+// Usage: yarn ts-node scripts/generate_user_space_activities_counters.ts
+async function main() {
+  const spaces = await db.queryAsync('SELECT id, name FROM spaces');
+
+  for (const index in spaces) {
+    console.log(
+      `Processing space #${spaces[index].id} (${spaces[index].name}) - ${index}/${spaces.length}`
```
ChaituVR, line 11: Incorrect index
```suggestion
      `Processing space #${spaces[index].id} (${spaces[index].name}) - ${index + 1}/${spaces.length}`
```

scripts/generate_user_space_activities_counters.ts
```diff
@@ -0,0 +1,38 @@
...
+// Usage: yarn ts-node scripts/generate_user_space_activities_counters.ts
+async function main() {
+  const spaces = await db.queryAsync('SELECT id, name FROM spaces');
+
+  for (const index in spaces) {
+    console.log(
+      `Processing space #${spaces[index].id} (${spaces[index].name}) - ${index + 1}/${spaces.length}`
+    );
+
+    const votesCountRes = await refreshVotesCount(spaces[index].id);
```
ChaituVR, line 14: But on delete-proposal we call it with an array `refreshVotesCount([msg.space]);`

ChaituVR (changes requested):

ChaituVR (approved):
tAck

src/writer/delete-space.ts
```diff
@@ -24,7 +25,12 @@ export async function action(body): Promise<void> {
   const space = msg.space;
 
   try {
-    await markSpaceAsDeleted(space);
+    const query = `
+      UPDATE spaces SET deleted = ? WHERE id = ? LIMIT 1;
+      DELETE FROM leaderboard WHERE space = ?;
+    `;
+
+    await db.queryAsync(query, [1, space, space]);
```
ChaituVR, line 33:
```suggestion
    const query = `
      UPDATE spaces SET deleted = 1 WHERE id = ? LIMIT 1;
      DELETE FROM leaderboard WHERE space = ?;
    `;

    await db.queryAsync(query, [space, space]);
```

PR #134, opened 2023-08-25
Redis store to rate limit

ChaituVR:
- Add `RATE_LIMIT_DATABASE_URL` env

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -72,5 +72,5 @@ export function getIp(req) {
     ''
   ).split(',');
 
-  return ips[0].trim();
+  return sha256(ips[0].trim()).substring(0, 10);
 }
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -72,5 +72,5 @@ export function getIp(req) {
     ''
   ).split(',');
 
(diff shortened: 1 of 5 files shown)


src/helpers/utils.ts
```diff
@@ -72,5 +72,5 @@ export function getIp(req) {
     ''
   ).split(',');
 
-  return ips[0].trim();
+  return sha256(ips[0].trim()).substring(0, 10);
```
wa0x6e, line 75: You should hash after retrieving that value. Editing the returned value is breaking current implementation

ChaituVR: Why 🤔 where are only using this as a key right 🤔

ChaituVR: Which one is breaking? how can i reproduce?

wa0x6e: Test are breaking. I think shutter is also using the unhashed IP

bonustrack: Good point, let's keep the function getIp unchanged and hash later in rateLimit file

bonustrack: We should do the same on score-api and hub

bonustrack (approved):
utACK

PR #137, opened 2023-08-29
chore(refactor): filter out client error from Sentry

### LINKED ISSUE #124: refactor: reject all promises with an error instance instead of string
Currently, we're using Promise for validation flow, and a rejection means error.

This also means that an invalid user input will be a promise rejection, on the same level as a server error, and there is currently no way to distinguish them apart, making the `catch()` function useless, as it's catching too much noise, and can not be used on Sentry.

From this doc (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject#description) it is recommended to reject the Promise with an instance of `Error`.

We should refactor the code, to have one (or more) new custom error class (UserInputError, etc...), so we can reject promises due to user input error with `Promise.reject(new UserInputError('Invalid schema'))`.

The ingestor which will catch that rejection can then check for the Error class, and return a 400 error for user related input, or return a 500 error and send it to Sentry on server error

wa0x6e:

Currently, `Promise.reject` are used for client error, and `throw` for server error. They're caught later up in the code, near the express router with a `try/catch`.

Current issue is that try/catch does not differentiate between user and server error, creating a lot of noise on Sentry

## 💊 Fixes / Solution

Fix #124

Inside the `catch` block, let's assume that when the argument is a `string`, it's `Promise.reject` due to client error and ignored from Sentry, and if it's an `Error` object, it's a server error and let's log it to Sentry.

This PR purpose is to filter out noise, so we can re-enable Sentry on this service

## 🚧 Changes

- Before sending errors to Sentry, filter and only send true errors, ignoring all Promise rejection.
- When failing due to client error, return a `400` error code. 
- Fix the `capture` function signature

## 🛠️ Tests

- POST some invalid result, it should return a 400 error code instead of 500 previously
- POST some valid result, it should return a 200 error

### DIFF
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -154,7 +154,9 @@ export default async function ingestor(req) {
   try {
     context = await writer[type].verify(legacyBody);
   } catch (e) {
-    capture(e);
+    if (typeof e !== 'string') {
+      capture(e);
+    }
     log.warn(`[ingestor] verify failed ${JSON.stringify(e)}`);
     return Promise.reject(e);
   }
@@ -193,7 +195,9 @@ export default async function ingestor(req) {
       receipt
     );
   } catch (e) {
-    capture(e);
+    if (typeof e !== 'string') {
+      capture(e);
+    }
     return Promise.reject(e);
   }
 
--- a/package.json
+++ b/package.json
@@ -9,7 +9,9 @@
     "build": "tsc",
     "dev": "nodemon src/index.ts",
     "start": "node build/src/index.js",
-    "test": "dotenv -e test/.env.test yarn jest"
+    "start:test": "dotenv -e test/.env.test yarn dev",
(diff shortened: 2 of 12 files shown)


ChaituVR (approved):
utAck

src/ingestor.ts
```diff
@@ -154,7 +154,9 @@ export default async function ingestor(req) {
   try {
     context = await writer[type].verify(legacyBody);
   } catch (e) {
-    capture(e);
+    if (typeof e !== 'string') {
+      capture(e);
+    }
```
ChaituVR, line 159: maybe better to handle this condition in capture itself? so we don't have to worry about adding it in the future? 

we may forgot adding it

wa0x6e: This is not systematic

PR #138, opened 2023-08-29
chore: instrument ingestor workflow

wa0x6e:
This PR add a instrument the ingestor function

## Changes

- Wrap the whole ingestor content inside a try/finally
- Move some variable definition outside the try block (so we can access them in finally)
- Start a timer, and end it in the `finally` block


src/ingestor.ts
```diff
@@ -12,208 +12,218 @@ import { storeMsg } from './helpers/highlight';
...
 
 const NAME = 'snapshot';
 const VERSION = '0.1.4';
 const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
 
 export default async function ingestor(req) {
-  const body = req.body;
+  const success = 0;
+  const type = '';
+  const network = '1';
```
ChaituVR, line 24: We are not changing this anywhere?

ChaituVR (approved):
utAck

src/ingestor.ts
```diff
@@ -12,208 +12,218 @@ import { storeMsg } from './helpers/highlight';
...
+import { timeIngestorProcess } from './helpers/metrics';
 
 const NAME = 'snapshot';
 const VERSION = '0.1.4';
 const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
 
 export default async function ingestor(req) {
-  const body = req.body;
+  const success = 0;
+  const type = '';
```
ChaituVR, line 23: We are not changing these anywhere? changes automatically?

wa0x6e: fixed, merge conflict gone wrong

PR #146, opened 2023-09-08
feat(proposal): add functionality to update proposal

Todmy:
Functionality for updating already created proposal. Details https://www.notion.so/snapshotlabs/Proposal-update-backend-c5f55c83f1ff4f3b82a273e96cb9c463?pvs=4
Lint errors here because of relation to PR https://github.com/snapshot-labs/snapshot.js/pull/895


src/ingestor.ts
```diff
@@ -171,13 +184,10 @@ export default async function ingestor(req) {
...
     let receipt;
     try {
-      const { address, sig, ...restBody } = body;
-      const ipfsBody = {
-        address,
-        sig,
-        hash: id,
-        ...restBody
-      };
+      const ipfsBody = writer[type].getIpfsBody
```
wa0x6e, line 187: What's the purpose of this check ? Afaik, none of the writer exposes a `getIpfsBody` function.

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,186 @@
...
+import { capture } from '@snapshot-labs/snapshot-sentry';
+import {
+  flaggedAddresses,
+  flaggedProposalTitleKeywords,
+  flaggedProposalBodyKeywords
+} from '../helpers/moderation';
+
+const scoreAPIUrl = process.env.SCORE_API_URL || 'https://score.snapshot.org';
+
+export function isChoicesValid({ type, choices }): boolean {
```
wa0x6e, line 17: What happens if some choices are being removed, while there are already existing votes with the removed choices ?

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,186 @@
...
+    space,
+    type: msg.payload.type || 'single-choice',
+    plugins,
+    title: msg.payload.name,
+    body: msg.payload.body,
+    discussion: msg.payload.discussion || '',
+    choices: JSON.stringify(msg.payload.choices)
+  };
+
+  const query = 'UPDATE proposals SET ? WHERE id = ?';
```
wa0x6e, line 159: Maybe add `LIMIT 1` ?

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,186 @@
...
+  const author = getAddress(body.address);
+  const updated = parseInt(msg.timestamp);
+  const metadata = msg.payload.metadata || {};
+  const plugins = JSON.stringify(metadata.plugins || {});
+
+  const proposal = {
+    ipfs,
+    author,
+    updated,
+    space,
```
wa0x6e, line 150: Isn't this always the same value before and after edit ? Maybe it can be ignored

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,186 @@
...
+  };
+
+  const ipfsBody = {
+    address,
+    sig,
+    hash: id,
+    originIpfs,
+    ...restBody,
+    data: restBody.data,
+    proposal: updatedProposal
```
wa0x6e, line 183: Isn't `body` already the updated proposal ?

wa0x6e: Why do we need to store `updatedProposal` ? the create proposal writer only store the body sent, and not the final proposal

src/writer/update-proposal.ts
wa0x6e: But what happens to existing votes with `==` which do no longer exist in the voting choices, but have been submitted ?

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,199 @@
...
+      if (validationName === 'any' || (validationName === 'basic' && !minScore)) {
+        isValid = true;
+      } else {
+        if (validationName === 'basic') {
+          validationParams.minScore = minScore;
+          validationParams.strategies = space.validation?.params?.strategies || space.strategies;
+        }
+
+        isValid = await snapshot.utils.validate(
+          validationName,
```
bonustrack, line 85: Why this? We dont need to check this to update a proposal, if author was able to propose he should be ablw to update we should only check author is same and proposal haven't started yet.

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,199 @@
...
+    ...(space.admins || []),
+    ...(space.moderators || [])
+  ].map(member => member.toLowerCase());
+
+  return members.includes(addressLC);
+}
+
+export async function checkAuthorization(address: string, space: any): Promise<string | undefined> {
+  const onlyAuthors = space.filters?.onlyMembers;
+  const isAuthorized = isAddressAuthorized(address, space);
```
bonustrack, line 65: We dont want this, only author of the proposal should be able to update proposal

src/writer/update-proposal.ts
bonustrack: It still better to add limit 1

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,199 @@
...
+  const updatedProposal = {
+    ...context,
+    ...message
+  };
+
+  const ipfsBody = {
+    address,
+    sig,
+    hash: id,
+    originIpfs,
```
bonustrack, line 193: We shouldn't change the format of the ipfs file

src/writer/update-proposal.ts
bonustrack: > What happens if some choices are being removed, while there are already existing votes with the removed choices ?

You can only update proposal before it start

src/writer/update-proposal.ts
wa0x6e: By saving the body twice (req.body, then updatedProposal), this increase the size of the saved JSON file, which has a limit of 1MB.

This may create edge-case, where updating a proposal fail because the pinning the IPFS file fail, because it's over 1MB, whereas the content itself did not increase

src/writer/update-proposal.ts
wa0x6e: Did not see any check related to that

bonustrack (changes requested):
Please always add description in PR and fix GitHub checks

src/writer/update-proposal.ts
Todmy: > Why do we need to store `updatedProposal` ? the create proposal writer only store the body sent, and not the final proposal

1MB is more than enough to store both entities. But anyway, what do you propose?
p.s.: you can find some considerations [here](https://discord.com/channels/955773041898573854/1149472664323379343/1149628031237885952)

wa0x6e: on proposal creation, it uses

```js
const { address, sig, ...restBody } = body;
const ipfsBody = {
  address,
  sig,
  hash: id,
  ...restBody
};
```

Why the same thing is not enough ? isn't `restBody` already containing the updated data ?

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,147 @@
...
+  const proposal = {
+    ipfs,
+    updated,
+    type: msg.payload.type || originalProposal.type,
+    plugins: plugins || originalProposal.plugins,
+    title: msg.payload.name || originalProposal.title,
+    body: msg.payload.body,
+    discussion: msg.payload.discussion,
+    choices: JSON.stringify(msg.payload.choices) || originalProposal.choices
+  };
```
wa0x6e, line 118: What does the UI looks like ? Isn't it coming from the same form as the create proposal form, in which case we should expect the full proposal, and don't really need to fallback on `originalProposal` ?

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,147 @@
...
+export function isAddressAuthorized(address: string, space: any): boolean {
+  const addressLC = address.toLowerCase();
+  const members = [
+    ...(space.members || []),
+    ...(space.admins || []),
+    ...(space.moderators || [])
+  ].map(member => member.toLowerCase());
+
+  return members.includes(addressLC);
+}
```
wa0x6e, line 57: This function is obsolete

src/writer/update-proposal.ts
Todmy: UI is not a part of this PR, you can check all details [here](https://www.notion.so/snapshotlabs/Proposal-update-backend-c5f55c83f1ff4f3b82a273e96cb9c463?pvs=4) and the task for UI is [here](https://www.notion.so/snapshotlabs/Allow-proposal-update-during-delay-period-289e6209cff5401f90fdb9c865efaf81?pvs=4)

wa0x6e: But what kind of payload are we expecting from UI ? The same as create proposal ? or if the user only edit the title, we only received the title ?

src/api.ts
```diff
@@ -21,7 +21,7 @@ router.post('/', async (req, res) => {
     const result = await typedData(req);
     return res.json(result);
   } catch (e) {
-    log.warn(`[ingestor] msg validation failed (typed data)`, e);
+    log.warn(`[ingestor] msg validation failed (typed data): ${e}`);
```
bonustrack, line 24: This wont work if the error is an object

src/writer/update-proposal.ts
Todmy: no, it will be a different payload. You can check it [here](https://github.com/snapshot-labs/snapshot.js/pull/895/files#diff-0de6060ecf3e628ed7a2ef320e36093f395de48084a4c2a4c2977588b6c0011dR143) and [here](https://github.com/snapshot-labs/snapshot.js/pull/895/files#diff-ff4da39e64c3be82c61a689b48eff52337efa570ec09f45be4f0f945f6f9097cR1)

src/api.ts
Todmy: checked with
```
const e = new Error("Cannot find module 'express'");
log.warn(`[ingestor] Wrong envelope format ${e}`);
```
and it outputs
```
warn: [ingestor] Wrong envelope format Error: Cannot find module 'express'
```
Can you provide an example of an error we need to handle? BTW, I have changed it because it wasn't work

src/writer/update-proposal.ts
Todmy: it will contain only the [update body](https://github.com/snapshot-labs/snapshot.js/pull/895/files#diff-0de6060ecf3e628ed7a2ef320e36093f395de48084a4c2a4c2977588b6c0011dR142), but not the whole body of the proposal.
> isn't restBody already containing the updated data ?

no, it contains only [fields](https://github.com/snapshot-labs/snapshot.js/pull/895/files#diff-0de6060ecf3e628ed7a2ef320e36093f395de48084a4c2a4c2977588b6c0011dR142) that are allowed to update

src/writer/update-proposal.ts
Todmy: do you propose to show only the data that was sent from UI to update a proposal?

src/api.ts
wa0x6e: Not sure if this apply in that case, but sometime, the error is a JSON-RPC error object

src/writer/update-proposal.ts
wa0x6e: So I guess the reason you're storing `proposal: updatedProposal` is to have the full proposal in the IPFS ?

src/api.ts
Todmy: followed the [same approach](https://github.com/snapshot-labs/snapshot-sequencer/pull/146/files#diff-769911c416ccf8514d8fd941ae0abe8fb5c606ade0c218e22151a5f5f9f3d700R47)

src/writer/update-proposal.ts
Todmy: Yes, + to have some kind of history with `originIpfs` key, that allows to track changes over IPFS storage

Todmy (commented):
fix(indentation): remove new line

src/ingestor.ts
```diff
@@ -178,6 +191,7 @@ export default async function ingestor(req) {
         hash: id,
         ...restBody
       };
+
```
Todmy, line 194:
```suggestion
```

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,113 @@
...
+  const metadata = msg.payload.metadata || {};
+  const plugins = JSON.stringify(metadata.plugins || {});
+
+  const proposal = {
+    ipfs,
+    updated,
+    type: msg.payload.type || originalProposal.type,
+    plugins: plugins || originalProposal.plugins,
+    title: msg.payload.name || originalProposal.title,
+    body: msg.payload.body,
```
wa0x6e, line 104: This will assume that on every update, you always send the `body`, updated or not, else if I update only the `title`, and only send the `title` in the payload, it'll wipe out the existing body.

Todmy: UI always needs to send all those fields, because we sign our data with [signTypedData](https://docs.ethers.org/v5/api/signer/#Signer-signTypedData). The fallback to `originalProposal` is used to restrict a user from setting some fields as empty. In this particular case I have assumed that `body` could be empty, but `title`, for example, cannot be

src/writer/update-proposal.ts
Todmy: I have removed those fields according to @bonustrack request. So I am resolving this discussion

src/writer/update-proposal.ts
wa0x6e: If UI always send the `title` field, why do you need to fallback to the original proposal ? Empty field prevention should already be done from the schema, isn't it ?

```json
"name": {
  "type": "string",
  "title": "name",
  "minLength": 1,
  "maxLength": 256
},
```

Todmy: because `Ajv` doesn't validate it.

wa0x6e:
package.json should also be updated to reflect the minimum required version, once snapshot.js version is released

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,115 @@
...
+    flaggedProposalTitleKeywords.some(keyword => proposalNameLC.includes(keyword))
+  );
+}
+
+export async function verify(body): Promise<any> {
+  const msg = jsonParse(body.msg);
+  const timestampNow = Math.floor(Date.now() / 1e3);
+
+  const schemaIsValid: any = snapshot.utils.validateSchema(
+    snapshot.schemas.updateProposal,
```
ChaituVR, line 53: Untitled 2

Todmy: yes, it's because we need to merge https://github.com/snapshot-labs/snapshot.js/pull/895 first

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,115 @@
...
+    msg.payload
+  );
+  if (schemaIsValid !== true) {
+    log.warn('[writer] Wrong proposal format', schemaIsValid);
+    return Promise.reject('wrong proposal format');
+  }
+
+  const proposal = await getProposal(msg.space, msg.payload.proposal);
+  if (!proposal) return Promise.reject('unknown proposal');
+  if (proposal.start < timestampNow) return Promise.reject('proposal already started');
```
ChaituVR, line 63: I think we should move `const timestampNow = Math.floor(Date.now() / 1e3);` above this line, just in case if there is some delay in `getProposal`

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,115 @@
...
+    space
+  });
+  if (spaceUpdateError) return Promise.reject(spaceUpdateError);
+
+  const hasScam = isScamDetected({
+    address: body.address,
+    name: msg.payload.name,
+    body: msg.payload.body
+  });
+  if (hasScam) return Promise.reject('scam proposal detected, contact support');
```
ChaituVR, line 89: I think we removed this check on proposal creation recently, maybe we should remove it here also?

src/writer/update-proposal.ts
```diff
@@ -0,0 +1,115 @@
...
+    switch (type) {
+      case 'basic':
+        return isEqual(['For', 'Against', 'Abstain'], choices);
+      default:
+        return choices.length > 0;
+    }
+  } else {
+    return false;
+  }
+}
```
ChaituVR, line 24: Maybe better to move this to utils and reuse on proposal creation too?

src/helpers/utils.ts
```diff
@@ -65,6 +66,19 @@ export function hasStrategyOverride(strategies: any[]) {
   return keywords.some(keyword => strategiesStr.includes(`"name":${keyword}`));
 }
 
+export function validateChoices({ type, choices }): boolean {
+  if (type && choices.length > 0) {
+    switch (type) {
+      case 'basic':
+        return isEqual(['For', 'Against', 'Abstain'], choices);
+      default:
+        return choices.length > 0;
```
ChaituVR, line 75:
```suggestion
      default:
        return true;
```

Todmy: is it possible to have 0 choices?

wa0x6e: No, since it’s nested inside an if with choices > 0

ChaituVR (approved):
tAck

PR #147, opened 2023-09-11
chore: update capture signature

wa0x6e:
This PR updates the `capture` call to use the shorter signature, added in  https://github.com/snapshot-labs/snapshot-sentry/releases/tag/v1.3.0


ChaituVR (approved):
utAck

PR #156, opened 2023-09-14
fix: remove banned keywords

bonustrack:


wa0x6e:
Need to fix test also

wa0x6e:
> Can you do that @wa0x6e ?

Will do

ChaituVR (approved):
utAck

PR #158, opened 2023-09-17
fix: instrument failed ingestor process

wa0x6e:
This PR instrument the failing ingestor process, logging which the error message, as well as the response time.

This instrumentation will give us metrics about how long has passed before the request is failing, with the given error.

Those metrics will let us optimize the code, to fail as earlier as possible


Todmy (approved):
utACK

PR #163, opened 2023-09-20
chore: fix pineapple error capture

wa0x6e:

New pineapple lib is returning the json-rpc error object nested in a `error` key, making the sentry automatic json-rpc parsing failing.

capture expect `{ code: '', message: ''}`, while pineapple is returning `{ error: { code: '', message: ''} }`.

## 💊 Fixes / Solution

When `error` property exist, capture it's value.

## 🚧 Changes

- capture the content of the `error` property when available

## 🛠️ Tests

- N/A

### DIFF
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -182,8 +182,8 @@ export default async function ingestor(req) {
         pin(ipfsBody, process.env.PINEAPPLE_URL),
         issueReceipt(body.sig)
       ]);
-    } catch (e) {
-      capture(e);
+    } catch (e: any) {
+      capture(e.error || e);
       return Promise.reject('pinning failed');
     }
     const ipfs = pinned.cid;


src/ingestor.ts
```diff
@@ -182,8 +182,8 @@ export default async function ingestor(req) {
         pin(ipfsBody, process.env.PINEAPPLE_URL),
         issueReceipt(body.sig)
       ]);
-    } catch (e) {
-      capture(e);
+    } catch (e: any) {
+      capture(e.error || e);
```
ChaituVR, line 186: I feel this should be handled inside the capture itself

wa0x6e: Let me see

wa0x6e:
Moved to https://github.com/snapshot-labs/snapshot-sentry/pull/10

PR #164, opened 2023-09-20
fix: ignore 429 from ingestor process error instrumentation

wa0x6e:
This PR exclude 429 errors from instrumented errors in the ingestor workflow.

429 errors are skewing the results, and creating a lot of noise


ChaituVR (approved):
utAck

PR #166, opened 2023-09-20
chore: customize the sourcemaps location path

wa0x6e:
Customize the sourcemaps location when uploading to Sentry


Todmy (approved):
utACK

PR #168, opened 2023-09-21
feat(flag): add action to flag proposal

Todmy:
Changes:
- add new action for flagging proposals

TODO:
- [ ] First [PR](https://github.com/snapshot-labs/snapshot-hub/pull/679) from snapshot-hub should be merged and DB migrated


src/writer/flag-proposal.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  return admins.includes(address.toLowerCase()) || mods.includes(address.toLowerCase());
+}
+
+export async function verify(body): Promise<any> {
+  const msg = jsonParse(body.msg);
+
+  const proposal = await getProposal(msg.space, msg.payload.proposal);
+  const space = await getSpace(msg.space);
+  if (!proposal) return Promise.reject('unknown proposal');
```
ChaituVR, line 17:
```suggestion
  if (!proposal) return Promise.reject('unknown proposal');
  const space = await getSpace(msg.space);
```

src/writer/flag-proposal.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const isAuthorizedToFlag = isAuthorized({
+    space,
+    address: body.address
+  });
+  if (!isAuthorizedToFlag) return Promise.reject('not authorized to flag proposal');
```
ChaituVR, line 23: Could we also check if the proposal is already flagged (as we already got the proposal info) better to reject it ❎ but not too important

Todmy: done on [line](https://github.com/snapshot-labs/snapshot-sequencer/pull/168/files#diff-3d4aaa652e292913a8c281d6a8d624a427f621dba27479cc2ad0940ee089c684R17)

ChaituVR (approved):
tAck

PR #169, opened 2023-09-21
chore: upgrade to pineapple  to 1.1.0

wa0x6e:
This PR upgrade pineapple.js to 1.1.0, which is adding support for retry on 504 errors


Todmy (approved):
utACK

PR #172, opened 2023-09-24
fix: prevent action replay

### LINKED ISSUE #27: feat: prevent replay of actions
Right now we accept duplicate requests, while this is fine in most cases, sometimes it can undo some actions, example if someone creates a proposal and then delete it, it is possible to replay create proposal action and get back the proposal,
It is possible to replay actions with timestamps in the last 3 days,

In such cases, we should check for duplicates (similar to vote action) and return an error

wa0x6e:
Fix #27 

Prevent action replay

- This check only if the given `id` already exists in the messages table
- Validate that hash is valid if passed in the body
- remove `IGNORE` from `storeMSg` query. The `isDuplicateMsg` check should prevent any INSERT errors.


bonustrack:
We should not rely on the signature to detect duplicate but the message id instead, there is multiple message with same sig, Safe always sign with "0x" as signature for example.

wa0x6e:
> We should not rely on the signature to detect duplicate but the message id instead, there is multiple message with same sig, Safe always sign with "0x" as signature for example.

fixed

ChaituVR:
Don't merge until https://github.com/snapshot-labs/snapshot-sequencer/pull/172/ else users may face an issue

wa0x6e:
What kind of issue?

wa0x6e:
`IGNORE` have been removed from the insert message query. This query should never fail, since we're now manually checking for duplicate

src/helpers/highlight.ts
```diff
@@ -16,3 +16,8 @@ export async function storeMsg(id, ipfs, address, version, timestamp, space, typ
     }
   ]);
 }
+
+export async function isDuplicateMsg(id: string) {
```
Todmy, line 20:
```suggestion
export async function doesMessageExist(id: string) Promise<boolean> {
```

src/ingestor.ts
```diff
@@ -101,6 +101,14 @@ export default async function ingestor(req) {
     const id = snapshot.utils.getHash(body.data);
     let payload = {};
 
+    if (body.hash && body.hash !== id) {
```
ChaituVR, line 104: In which case do we pass `hash` in the body?

wa0x6e: I have no idea, seems like it’s optional

ChaituVR: hmm don't see it any where else, maybe we can remove this if condition

src/helpers/highlight.ts
ChaituVR: what did we decide on this 😄 

GPT says:
image

Todmy: I propose this name, based on the function functionality. It says whether a message with a passed `id` exists or not. `isDuplicateMsg` sounds more like: "check duplicates somewhere", but it's not duplicated, because we didn't save it.
p.s.: I don't think that we need to spend much time on this.

PR #173, opened 2023-09-25
feat(flag): adds endpoint to flag proposals/spaces

Todmy:
Changes:
- adds new endpoint POST `/flag` that is secured with header `secret`. This endpoint could be used only internally 
- endpoint accepts body:
```json
{
    "type": "space | proposal",
    "value": "<space id> | <proposal id>",
    "action": "flag | verify"
}
```
where `flag` could be used with `space` and `proposal`, but `verify` only with `space`

TODO:
- [x] First should be merged [PR in snapshot-hub](https://github.com/snapshot-labs/snapshot-hub/pull/682)


src/api.ts
```diff
@@ -67,4 +68,15 @@ router.get('/spaces/:key/poke', async (req, res) => {
   }
 });
 
+router.post('/flag', verifyAuth, async (req, res) => {
+  const { type, value, action } = req.body;
+
+  try {
+    await flagEntity({ type, value, action });
+    return res.json({ success: true });
```
wa0x6e, line 76: This action always return true, whether the space/proposal exist or not. Maybe use the `affectedRows` from the query result to return whether the sql query has succeed or not ?

wa0x6e (approved):
tAck

src/api.ts
Todmy: We will use this endpoint inside the laser. And previously laser hasn’t handled this scenario before (you were able to insert the wrong values). I just followed the previous approach we had

src/helpers/utils.ts
```diff
@@ -90,3 +90,15 @@ export function getIp(req) {
 
   return ips[0].trim();
 }
+
+export function verifyAuth(req, res, next) {
+  const auth = req.headers.secret || '';
+  const authHash = sha256(auth);
+  const secretHash = process.env.AUTH_SECRET;
```
ChaituVR, line 97: Can we rename `AUTH_SECRET`  so we remember it is related to laser or do we plan to use it for more things in the future?

.env.example
```diff
@@ -11,3 +11,4 @@ SCORE_API_URL=https://score.snapshot.org
 # If you need unlimted access to score-api, use `https://score.snapshot.org?apiKey=...`
 RATE_LIMIT_DATABASE_URL= # optional
 BROVIDER_URL=https://rpc.snapshot.org # optional
+AUTH_SECRET=1dfd84a695705665668c260222ded178d1f1d62d251d7bee8148428dac6d0487
```
ChaituVR, line 14:
```suggestion
# Secret for laser
AUTH_SECRET=1dfd84a695705665668c260222ded178d1f1d62d251d7bee8148428dac6d0487 # optional
```

src/helpers/moderation.ts
```diff
@@ -32,3 +33,28 @@ export default async function run() {
   await snapshot.utils.sleep(20e3);
   run();
 }
+
+export function flagEntity({ type, action, value }) {
+  if (!type || !value || !action)
+    throw new Error(`missing params. 'type', 'value' and 'action' required`);
```
ChaituVR, line 39:
```suggestion
  if (!type || !action || !value )
    throw new Error(`missing params. 'type', 'action' and 'value' required`);
```

ChaituVR (approved):
utAck

src/helpers/moderation.ts
```diff
@@ -32,3 +33,28 @@ export default async function run() {
   await snapshot.utils.sleep(20e3);
   run();
 }
+
+export function flagEntity({ type, action, value }) {
+  if (!type || !action || !value )
```
ChaituVR, line 38:
```suggestion
  if (!type || !action || !value)
```

PR #177, opened 2023-09-30
chore: migrate from cross-fetch to node-fetch

wa0x6e:
This PR migrates the fetch library from `cross-fech` to `node-fetch`.

This code is node only, and does not require browser support from cross-fetch.


ChaituVR:
We will add timeouts and keep-alive in different PR?

wa0x6e:
Yes sir

ChaituVR (approved):
utAck

PR #179, opened 2023-09-30
feat: add `keep-alive` and `timeout` to moderation fetching

wa0x6e:

Current fetch connection to sidekick, retrieving the moderation data, is not benefitting from keep-alive connections, which sometimes, will cause a 504 error.

## 💊 Fixes / Solution

Fix SNAPSHOT-SEQUENCER-E

## 🚧 Changes

- Add `keep-alive` and `timeout` to connection to sidekick

## 🛠️ Tests

- N/A


ChaituVR (approved):
utAck

PR #180, opened 2023-10-01
fix: remove`IGNORE` in sql queries

wa0x6e:

Following #172 

Most of the queries with `IGNORE` are now unnecessary, as duplication is now manually checked before triggering the query, and use of `IGNORE` is masking errors.

Query with `ON DUPLICATE SET ?` are also obsolete, since no duplicate are allowed.

## 💊 Fixes / Solution

- Remove `IGNORE` from all writers's SQL query.
- In the ingestor try/catch, catch sql duplicate error (just in case, should never happen)

## Notes 

Waiting for #172 to be merged before


src/ingestor.ts
```diff
@@ -226,10 +226,15 @@ export default async function ingestor(req) {
...
-    } catch (e) {
+    } catch (e: any) {
       if (typeof e !== 'string') {
         capture(e);
       }
+
+      if (e.errno === 1062) {
+        return Promise.reject('duplicate message');
+      }
+
```
ChaituVR, line 237: I think we should place it before capture, to avoid these errors on sentry. or we want to know them for now?

wa0x6e: PR is to be merged after the duplication check PR, after when, this error should not happen. We capture the error in case of.

src/writer/proposal.ts
```diff
@@ -191,6 +191,7 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
     type: msg.payload.type || 'single-choice',
     strategies,
     plugins,
+    delegation: 0,
```
ChaituVR, line 194: we can remove this now

src/writer/statement.ts
```diff
@@ -29,9 +29,7 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
     created
   };
 
-  const query =
-    'INSERT IGNORE INTO statements SET ? ON DUPLICATE KEY UPDATE ipfs = ?, about = ?, statement = ?, updated = ?';
-  const params = [item, item.ipfs, item.about, item.statement, item.created];
+  const query = 'INSERT INTO statements SET ?';
```
ChaituVR, line 32: This should update existing statement (in case of duplicate key)

ChaituVR (changes requested):

src/writer/statement.ts
wa0x6e: In what scenario there can be duplicate key ?

In the prevent action replay PR, we ensured that the ID does not exist

ChaituVR: Yea yea this for `statements` table here primary key is `PRIMARY KEY (delegate, space),` if we receive a statement from same delegate we update existing record

wa0x6e: Mhhh. did not see those `PRIMARY KEY`. PR need more works to take these into account.

wa0x6e:
Let's also wait for #216 to make sure no errors from duplicate keys can happen anymore

src/ingestor.ts
ChaituVR: We don't need this anymore correct?

wa0x6e: I think this need #216

wa0x6e: Check removed, since should not happen anymore since merging of #216

ChaituVR:
I think we should fix tests
> expect(received).rejects.toEqual(expected) // deep equality

    Expected: "duplicate message"
    Received: [Error: ER_DUP_ENTRY: Duplicate entry '0x76658bb0d8e5dec677159ae78797ea8bb16eede4ff09fa7c4d5d71010dcccd' for key 'proposals.PRIMARY']

      197 |     await expect(ingestor(proposalRequest)).resolves.toHaveProperty('id');
      198 |     await db.queryAsync('DELETE from snapshot_sequencer_test.messages');
    > 199 |     await expect(ingestor(proposalRequest)).rejects.toEqual('duplicate message');
          |                                                     ^
      200 |   });

wa0x6e:
> I think we should fix tests
> 
> > expect(received).rejects.toEqual(expected) // deep equality
> 
> ```
> Expected: "duplicate message"
> Received: [Error: ER_DUP_ENTRY: Duplicate entry '0x76658bb0d8e5dec677159ae78797ea8bb16eede4ff09fa7c4d5d71010dcccd' for key 'proposals.PRIMARY']
> 
>   197 |     await expect(ingestor(proposalRequest)).resolves.toHaveProperty('id');
>   198 |     await db.queryAsync('DELETE from snapshot_sequencer_test.messages');
> > 199 |     await expect(ingestor(proposalRequest)).rejects.toEqual('duplicate message');
>       |                                                     ^
>   200 |   });
> ```

Tests are failing, but not this one. This one is from duplicate message, not the same duplicate as in parallel

wa0x6e:
There's still a `IGNORE` in `addOrUpdateSpace()`, but don't remember why it was left, maybe a miss, or may be a valid reason, can not remember since this PR is quite old

ChaituVR (approved):
tAck

In worst case scenario it will throw some error, we should look out for errors

PR #181, opened 2023-10-01
fix: remove unused delegation field

wa0x6e:
The `delegation` column in the proposals table is unused, and should be removed.

Should run this query to drop the column 

```sql
ALTER TABLE proposals DROP COLUMN delegation;
```


ChaituVR (approved):
utAck

test/schema.sql
```diff
@@ -32,7 +32,6 @@ CREATE TABLE proposals (
   choices JSON NOT NULL,
   start INT(11) NOT NULL,
   end INT(11) NOT NULL,
-  delegation INT(1) NOT NULL,
```
ChaituVR, line 35: Is there some other way to use schema from hub 🙈  we didn't add many fields here from there recently 🙈

wa0x6e: What you have in mind ? github actions to sync the file with hub ?

ChaituVR: Excellent Idea! I think we can check for changes on each pull request sequencer and sync with the hub

PR #189, opened 2023-10-04
refactor: Use ajv format instead of getAddress on all actions

ChaituVR:
Changes:
- Check `format` of `address` and `from` using ajv format https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L44-L52
- `getAddress` on all actions won't be useful anymore


wa0x6e:
We're losing the case formatting, does it matter ?

```js
getAddress("0x8ba1f109551bd432803012645ac136ddd64dba72");
// '0x8ba1f109551bD432803012645Ac136ddd64DBA72'
```

src/writer/alias.ts
```diff
@@ -9,7 +9,7 @@ export async function action(message, ipfs, receipt, id): Promise<void> {
   const params = {
     id,
     ipfs,
-    address: getAddress(message.from),
+    address: message.from,
     alias: getAddress(message.alias),
```
wa0x6e, line 13: Why leave `getAddress` on this one ?

wa0x6e: By removing this last one, we could also remove `@ethersproject/address` from package dependency

ChaituVR:
> We're losing the case formatting, does it matter ?
> 
> ```js
> getAddress("0x8ba1f109551bd432803012645ac136ddd64dba72");
> // '0x8ba1f109551bD432803012645Ac136ddd64DBA72'
> ```

Yes yes. Address should be always checksummed, `address` format checks if address is checksummed or not
It will reject the request if address is not checksummed, so we don't need to run `getAddress` on all actions

wa0x6e:
> We're losing the case formatting, does it matter ?
> 
> ```js
> getAddress("0x8ba1f109551bd432803012645ac136ddd64dba72");
> // '0x8ba1f109551bD432803012645Ac136ddd64DBA72'
> ```

In case formatting does matter, we could use `getAddress` only on the writer actions itself, before inserting in the DB, and keep ajv for validation only

src/writer/alias.ts
ChaituVR: Agree. Hehe will give it a try

Todmy (approved):
tACK

wa0x6e:
Should this one also wait on https://github.com/snapshot-labs/snapshot-sequencer/pull/192 ?

ChaituVR:
Updated PR

wa0x6e (approved):
tAck

PR #191, opened 2023-10-05
chore: make logging consistent

wa0x6e:
- Remove redundant `console.log`
- Move the last few `log.error` to `log.warn`

### DIFF
--- a/src/helpers/shutter.ts
+++ b/src/helpers/shutter.ts
@@ -107,7 +107,7 @@ export async function setProposalKey(params) {
     log.info(`[shutter] proposal scores updated for ${proposalId}`);
   } catch (e) {
     capture(e);
-    log.error(`[shutter] setProposalKey failed ${JSON.stringify(e)}`);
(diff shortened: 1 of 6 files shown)


src/helpers/shutter.ts
```diff
@@ -107,7 +107,7 @@ export async function setProposalKey(params) {
     log.info(`[shutter] proposal scores updated for ${proposalId}`);
   } catch (e) {
     capture(e);
-    log.error(`[shutter] setProposalKey failed ${JSON.stringify(e)}`);
```
ChaituVR, line 110: Maybe we can leave it an error. Slightly more important

ChaituVR (approved):
utAck

src/helpers/shutter.ts
wa0x6e: In that case, we should also move all `log.warn` to `log.error`, unless they're validation error (in ingestor).

wa0x6e: Let's not forget we also capture the error on Sentry, to that log should be gone

PR #192, opened 2023-10-05
fix: Use alias schema

ChaituVR:
Merge after updating snapshot.js version https://github.com/snapshot-labs/snapshot.js/pull/913

### DIFF
--- a/src/writer/alias.ts
+++ b/src/writer/alias.ts
@@ -9,7 +9,7 @@ export async function verify(message): Promise<any> {
   const schemaIsValid = snapshot.utils.validateSchema(snapshot.schemas.alias, msg.payload);
   if (schemaIsValid !== true) {
     log.warn(`[writer] Wrong alias format ${JSON.stringify(schemaIsValid)}`);
-    return Promise.reject('wrong statement format');
+    return Promise.reject('wrong alias format');
   }
   return message.from !== msg.payload.alias;
 }
--- a/src/writer/alias.ts
+++ b/src/writer/alias.ts
@@ -8,7 +8,7 @@ export async function verify(message): Promise<any> {
   const msg = jsonParse(message.msg, {});
   const schemaIsValid = snapshot.utils.validateSchema(snapshot.schemas.alias, msg.payload);
   if (schemaIsValid !== true) {
-    log.warn(`[writer] Wrong statement format ${JSON.stringify(schemaIsValid)}`);
+    log.warn(`[writer] Wrong alias format ${JSON.stringify(schemaIsValid)}`);
     return Promise.reject('wrong statement format');
   }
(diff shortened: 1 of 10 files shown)


src/writer/alias.ts
```diff
@@ -1,8 +1,17 @@
...
+import { jsonParse } from '../helpers/utils';
+import log from '../helpers/log';
 
 export async function verify(message): Promise<any> {
-  return message.from !== message.alias;
+  const msg = jsonParse(message.msg, {});
+  const schemaIsValid = snapshot.utils.validateSchema(snapshot.schemas.alias, msg.payload);
+  if (schemaIsValid !== true) {
+    log.warn(`[writer] Wrong statement format ${JSON.stringify(schemaIsValid)}`);
+    return Promise.reject('wrong statement format');
```
wa0x6e, line 12:
```suggestion
    return Promise.reject('wrong alias format');
```

src/writer/alias.ts
```diff
@@ -1,8 +1,17 @@
...
 import db from '../helpers/mysql';
+import { jsonParse } from '../helpers/utils';
+import log from '../helpers/log';
 
 export async function verify(message): Promise<any> {
-  return message.from !== message.alias;
+  const msg = jsonParse(message.msg, {});
+  const schemaIsValid = snapshot.utils.validateSchema(snapshot.schemas.alias, msg.payload);
+  if (schemaIsValid !== true) {
+    log.warn(`[writer] Wrong statement format ${JSON.stringify(schemaIsValid)}`);
```
wa0x6e, line 11:
```suggestion
    log.warn(`[writer] Wrong alias format ${JSON.stringify(schemaIsValid)}`);
```

wa0x6e (changes requested):

src/writer/alias.ts
```diff
@@ -1,8 +1,17 @@
...
+    return Promise.reject('wrong statement format');
+  }
+  return message.from !== msg.payload.alias;
```
wa0x6e, line 14:
```suggestion
  if (message.from === msg.payload.alias) {
    return Promise.reject('alias cannot be the same as the address');
  }
```

wa0x6e (changes requested):
`verify` expect a Promise.rejection on error

wa0x6e:
I've also commit some tests

src/writer/alias.ts
ChaituVR:
```suggestion
  if (message.from === msg.payload.alias) {
	    return Promise.reject('alias cannot be the same as the address');
	}
  return true;
```

wa0x6e:
Lint error coming from suggested changes 🙈

wa0x6e (approved):
tAck

PR #194, opened 2023-10-06
fix: check for duplicate vote in verify

wa0x6e:
If a vote already exist, we check that the new vote is more recent that the existing one before inserting, in the vote's `action` function.

We should also do the same in the `verify` function, to avoid a call to get voting power.

This kind of error is currently the slowest one:

Screenshot 2023-10-06 at 06 55 55


src/writer/vote.ts
```diff
@@ -48,6 +48,17 @@ export async function verify(body): Promise<any> {
...
+  // Check if voter already voted
+  const votes = await db.queryAsync(
+    'SELECT created FROM votes WHERE voter = ? AND proposal = ? AND space = ? ORDER BY created DESC LIMIT 1',
+    [body.address, msg.payload.proposal, msg.space]
+  );
+
+  // Reject vote with later timestamp
+  if (votes[0]?.created > msgTs) {
+    return Promise.reject('already voted at later time');
+  }
```
ChaituVR, line 60: It is already checked in action right https://github.com/snapshot-labs/snapshot-sequencer/blob/use-alias-schema/src/writer/vote.ts#L148 ? maybe we should remove that if we already check it in verify

wa0x6e: I think the check inside `action` is to there in case some other threads/container are processing the same exact request, and is still needed, since we can't guarantee that the check inside `verify` is valid.

ChaituVR: mm better to handle parallel requests in requestDeduplicator, because this can cause duplicate query and code, meaning additional time to vote for users 🙈

wa0x6e: We should move all inside `verify`, after implementing requestDeduplicator #202

ChaituVR: mmm better to install deduplication first

wa0x6e:
Superseded by #216 , should still move the check from `action` to `verify`

PR #197, opened 2023-10-08
feat: add script to hibernate spaces

wa0x6e:
Pitch: https://github.com/snapshot-labs/pitches/issues/65

- Update schema for new `hibernated` column
- Add a script to hibernate spaces (or see https://github.com/snapshot-labs/snapshot-sequencer/pull/197#issuecomment-1808168112 for the direct SQL query)
- Auto reactivate space on settings update (will require https://github.com/snapshot-labs/snapshot-sequencer/pull/240)

## Usage

- `yarn ts-node scripts/hibernate.ts run` will hibernate inactive spaces
- `yarn ts-node scripts/hibernate.ts preview` will show the number of spaces, eligible for hibernation

## Note 

This is PR is only about supporting the new `hibernated` column, as well as putting spaces to hibernation, and does not handled any kind validation (which requires coordination with UI, and come in a later PR)


src/helpers/hibernation.ts
```diff
@@ -0,0 +1,10 @@
+import db from './mysql';
+
+export default function hibernate(id: string, action: string) {
+  if (!id || !action) throw new Error(`missing params. 'id' and 'action' required`);
+  if (!['hibernate', 'reactivate'].includes(action)) throw new Error('invalid action');
+
+  const query = `UPDATE spaces SET hibernating = ? WHERE id = ? LIMIT 1`;
```
bonustrack, line 7: We should use `hibernated` name, similar than `created` or `updated`
```suggestion
  const query = `UPDATE spaces SET hibernated = ? WHERE id = ? LIMIT 1`;
```

test/schema.sql
```diff
@@ -5,12 +5,15 @@ CREATE TABLE spaces (
   verified INT NOT NULL DEFAULT '0',
   deleted INT NOT NULL DEFAULT '0',
   flagged INT NOT NULL DEFAULT '0',
+  hibernated INT NOT NULL DEFAULT '0',
+  created BIGINT NOT NULL,
```
ChaituVR, line 9:
```suggestion
```

wa0x6e:
Script has been stripped down to hibernate only spaces inactive more than 1 year ago, and SQL simple enough to not warrant a script.ts anymore.

Running the following SQL query will hibernate the spaces

```sql
WITH toHibernate AS (
      WITH data AS (
        SELECT
        id,
        (SELECT MAX(end) FROM proposals WHERE space = spaces.id LIMIT 1) AS lastProposalEndDate
        FROM spaces
        WHERE hibernated = 0
      )

      SELECT
        id, lastProposalEndDate
        FROM data
        WHERE
        # Filtering out spaces that have not been active in the past year
        FROM_UNIXTIME(lastProposalEndDate) < DATE_SUB(CURRENT_DATE, INTERVAL 1 YEAR)
    )
    UPDATE spaces SET hibernated = 1 where id IN (id)
```

ChaituVR (approved):
utAck

PR #200, opened 2023-10-09
feat: Use only sequencer DB for messages

ChaituVR:
Related to https://www.notion.so/snapshotlabs/Move-all-messages-to-sequencer-f7843f0b1f3049b79f87647ca652d181
To use only sequencer DB instead of both hub and sequencer DB like before

- [x] Merge only once sequencer DB is successfully synced with hub DB
- [x] Merge after merging https://github.com/snapshot-labs/snapshot-hub/pull/696


Todmy (approved):
utACK

PR #203, opened 2023-10-09
fix: use moderation data from hub

wa0x6e:

The rateLimit module is still using flagged and verified spaces from sidekick, which is now obsolete

## 💊 Fixes / Solution

Use flagged and verified spaces from hub directly

## 🚧 Changes

- Change the rateLimit code to fetch space's `verified` and `flagged` status directly from the space object 
- the `getSpaceLimits` function inside helpers/limit.ts now take a Space object, instead of spaceId
- When using `getSpace()`, check after that space is set before using

## 🛠️ Tests

- Tests have been added


ChaituVR (approved):
tAck

PR #204, opened 2023-10-10
chore: add script to setup test db

wa0x6e:
- Add a new `test:setup` yarn task to setup the test database. This task will run the commands to create the test database, and import the schema.
- Warn before booting the test suite is the test database is missing


ChaituVR (approved):
utAck

package.json
```diff
@@ -10,6 +10,7 @@
     "dev": "nodemon src/index.ts",
     "start": "node build/src/index.js",
     "start:test": "dotenv -e test/.env.test yarn dev",
+    "test:setup": "./test/setup.sh",
```
Todmy, line 13: what is the process of running it? Should I define some variables?
Screenshot 2023-10-10 at 11 56 45

wa0x6e: Is your mysql database running locally ? Do you have the `mysql` cli installed ?

wa0x6e: > yes, it runs locally. No, I don't have `mysql` cli - maybe we should add it to devDependencies if it's required?

The `mysql` cli should be shipped with the mysql install, and can not be installed with npm

Todmy (approved):
utACK

wa0x6e:
Let me find another way to create the database

Todmy:
> Let me find another way to create the database

I think it's not too critical. At least the script provides an example

wa0x6e:
Script have been completely refactor to use 100% nodejs code, instead of shell command.

This will fix the `mysql` cli requirement issue.

Todmy (approved):
tACK

Todmy:
thank you 🙏 it's awesome

ChaituVR (approved):
utAck

PR #205, opened 2023-10-10
fix: Check limit with verified and flagged properties

ChaituVR:
Right now it is not considering latest change to store verified and flagged properties from database


ChaituVR:
Already fixed in #203

PR #206, opened 2023-10-16
feat: Add limit for follows per user

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot/issues/4286

We now limit users to have the follows of 50 spaces. They won't be able to follow more spaces than that

### DIFF
--- a/src/writer/follow.ts
+++ b/src/writer/follow.ts
@@ -1,6 +1,24 @@
+import { FOLLOWS_LIMIT_PER_USER } from '../helpers/limits';
 import db from '../helpers/mysql';
 
-export async function verify(): Promise<any> {
+export const getFollowsCount = async follower => {
+  const query = `
+  SELECT
+    COUNT(*) AS followsCount
+  FROM
+    follows
+  WHERE
+    follower = ?
+  `;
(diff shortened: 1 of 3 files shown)


src/writer/follow.ts
```diff
@@ -1,6 +1,24 @@
...
+    follows
+  WHERE
+    follower = ?
+  `;
+  return await db.queryAsync(query, follower);
+};
+
+export async function verify(message): Promise<any> {
+  const follower = message.from;
+  const [{ followsCount }] = await getFollowsCount(follower);
```
wa0x6e, line 18: Wouldn't that deconstructor throw an error if the array is empty ?

wa0x6e (approved):
tACK

src/writer/follow.ts
```diff
@@ -1,6 +1,24 @@
+import { FOLLOWS_LIMIT_PER_USER } from '../helpers/limits';
 import db from '../helpers/mysql';
 
-export async function verify(): Promise<any> {
+export const getFollowsCount = async follower => {
```
bonustrack, line 4: Based on this function name it should return a number and not an array of object

PR #207, opened 2023-10-17
chore: setup db on test run

wa0x6e:
- Auto-run `yarn test:setup` before `yarn:test`
- Remove unused `cross-fetch` dependencies
- Remove `supertest` dependency, replace by simpler `node-fetch`


ChaituVR (approved):
tAck

PR #215, opened 2023-10-20
fix: enforce proposal `end` date to be in the future

### LINKED ISSUE #214: Allow proposals only when `end` param is greater than current timestamp
Related to https://github.com/snapshot-labs/snapshot/issues/3339

wa0x6e:
Fix #214 

Enforce proposal `end` property to be in the future

### DIFF
--- a/src/writer/proposal.ts
+++ b/src/writer/proposal.ts
@@ -47,7 +47,7 @@ export async function verify(body): Promise<any> {
   }
 
   const tsInt = (Date.now() / 1e3).toFixed();
-  if (msg.payload.end < tsInt) {
+  if (msg.payload.end <= tsInt) {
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
utAck

src/writer/proposal.ts
```diff
@@ -46,6 +46,11 @@ export async function verify(body): Promise<any> {
     return Promise.reject('wrong proposal format');
   }
 
+  const tsInt = (Date.now() / 1e3).toFixed();
+  if (msg.payload.end < tsInt) {
```
ChaituVR, line 50: I feel like we can make it "Not to be in past"
```suggestion
  if (msg.payload.end <= tsInt) {
```

PR #216, opened 2023-10-20
feat: prevent duplicate requests

### LINKED ISSUE #195: Error: ER_DUP_ENTRY: Duplicate entry '0x56faebdfc15c9de29191e4b0e624e0382fcea1d5b9085ac61d1d210eb62740' ...
Sentry Issue: [SNAPSHOT-SEQUENCER-R](https://snapshot-labs.sentry.io/issues/4416425857/?referrer=github_integration)

```
Error: ER_DUP_ENTRY: Duplicate entry '0x56faebdfc15c9de29191e4b0e624e0382fcea1d5b9085ac61d1d210eb62740' for key 'messages.PRIMARY'
  File "/workspace/build/src/helpers/highlight.js", line 10, in storeMsg
    await mysql_1.default.queryAsync(query, [
  File "/workspace/build/src/ingestor.js", line 231, in ingestor
    await (0, highlight_1.storeMsg)(id, ipfs, body.address, msg.version, msg.timestamp, msg.space || '', msg.type, body.sig, receipt);
  File "/workspace/build/src/api.js", line 25, in <anonymous>
    const result = await (0, ingestor_1.default)(req);
...
(19 additional frame(s) were not displayed)

Error: ER_DUP_ENTRY: Duplicate entry '0x56faebdfc15c9de29191e4b0e624e0382fcea1d5b9085ac61d1d210eb62740' for key 'messages.PRIMARY'
  File "/workspace/build/src/helpers/highlight.js", line 10, in storeMsg
    await mysql_1.default.queryAsync(query, [
  File "/workspace/build/src/ingestor.js", line 231, in ingestor
    await (0, highlight_1.storeMsg)(id, ipfs, body.address, msg.version, msg.timestamp, msg.space || '', msg.type, body.sig, receipt);
  File "/workspace/build/src/api.js", line 25, in <anonymous>
    const result = await (0, ingestor_1.default)(req);
...
(19 additional frame(s) were not displayed)
```

### LINKED ISSUE #202: feat: install requestDeduplicator
Fix #195

### LINKED ISSUE #217: bug: race condition in limit check
When submitting something with a limit, limit is checked at the start of the `verify` action, but multiple requests submitted at the same time can bypass the limit check

wa0x6e:

We have a check preventing action replay, by checking if the message already exists in the database, but there is an edge case, where that check fail if multiple identical requests are sent at the same time, and none of the messages are saved in the database yet.

This scenario results in the ingestor not considering the requests as duplicate, will process the action, but will fail when saving the messages, with the `ER_DUP_ENTRY: Duplicate entry` error, or in the writer's `action` function (depending on the action type)

## 💊 Fixes / Solution

Fix #202 
Fix #217 
Fix #195
Fix [SNAPSHOT-SEQUENCER-R](https://snapshot-labs.sentry.io/issues/4416425857/?referrer=github_integration)

- Have a temporary central database of actions currently in the processing queue.
- When a request start, we add the request signature in the DB, and remove it on completion.
- On each new request, check if its signature is present in the DB, and return a 429 error if true, else process it.

## 🚧 Changes

- Add a new middleware `duplicateRequestPreventor`, to prevent duplicate requests

## 🛠️ Tests

- [x] Add tests
- [x] Waiting for all other PR to be merged to decrease changes files count

## Notes

<strike> This middleware still does not handle the edge case where:

- User send request A (which usually take 1-2s to complete)
- User immediately abort the request A (key is released, can send same request again)
- User resend request A

Both request A are processed in this case </strike>


src/helpers/rateLimit.ts
```diff
@@ -42,3 +48,26 @@ export default rateLimit({
...
+  if (!client) {
+    return next();
+  }
+
+  const value = hashedBody(req);
+  const [duplicate] = await client
+    .multi()
+    .SISMEMBER(DUPLICATOR_SET_KEY, value)
+    .SADD(DUPLICATOR_SET_KEY, value)
+    .exec();
```
ChaituVR, line 62: All to redis? will be costly? maybe fine

ChaituVR (commented):
Looks good, will test tmrw

src/helpers/rateLimit.ts
wa0x6e: Saving to redis should be fine, it's made for this. What may be costly is sha256 the whole body, don't know its cost yet

src/helpers/rateLimit.ts
```diff
@@ -42,3 +48,26 @@ export default rateLimit({
       })
     : undefined
 });
+
+export async function duplicateRequestLimit(req: Request, res: Response, next: NextFunction) {
```
ChaituVR, line 52: Could we move this to a separate file? It is defn. not related to the rate-limit 😄

ChaituVR (changes requested):
Oct-24-2023 23-57-05

So i added a timeout on 
```js
router.get('/', async (req, res) => {
  const commit = process.env.COMMIT_HASH || '';
  const v = commit ? `${version}#${commit.substr(0, 7)}` : version;
  // wait for 10 seconds
  await new Promise(resolve => setTimeout(resolve, 10000));
  return res.json({
    name,
    network,
    version: v,
    relayer: relayer.address
  });
});
```


Weirdly if I open the same request on different tabs all of them are waiting for response, than returning an error, where on postman it return error, which is correct 🤔

src/helpers/rateLimit.ts
wa0x6e: Refactored to use `body.sig` instead, will improve performance

wa0x6e:
> Oct-24-2023 23-57-05 [ Oct-24-2023 23-57-05 ](https://user-images.githubusercontent.com/15967809/277767416-2e6fbbd5-e16a-4e71-8232-7c894d9afbb1.gif) [ ](https://user-images.githubusercontent.com/15967809/277767416-2e6fbbd5-e16a-4e71-8232-7c894d9afbb1.gif)
> 
> So i added a timeout on
> 
> ```js
> router.get('/', async (req, res) => {
>   const commit = process.env.COMMIT_HASH || '';
>   const v = commit ? `${version}#${commit.substr(0, 7)}` : version;
>   // wait for 10 seconds
>   await new Promise(resolve => setTimeout(resolve, 10000));
>   return res.json({
>     name,
>     network,
>     version: v,
>     relayer: relayer.address
>   });
> });
> ```
> 
> Weirdly if I open the same request on different tabs all of them are waiting for response, than returning an error, where on postman it return error, which is correct 🤔

Something going with Chrome sessions. It returns the correct error when using a different session (private navigation).

src/helpers/duplicateRequestPreventor.ts
```diff
@@ -0,0 +1,39 @@
+import type { Request, Response, NextFunction } from 'express';
+import { sendError } from './utils';
+import redisClient from './redis';
```
ChaituVR, line 3: Why we are using redis ? other services don't use Redis. unlike other services, we use single instance for sequencer no?

ChaituVR (changes requested):

wa0x6e:
Don’t remember exactly why, maybe sequencer was on multiple instances then

src/helpers/duplicateRequestPreventor.ts
wa0x6e: Don’t remember the exact reason, maybe sequencer was running on multiple instances then

wa0x6e: Refactor to remove Redis dependency.

I think Redis was for better access to the duplicator queue from the tests, which are now refactored as integration test instead of e2e

src/helpers/duplicateRequestPreventor.ts
```diff
@@ -0,0 +1,24 @@
...
+const hashedBody = (req: Request): string => JSON.stringify(req.body?.sig || {});
+
+export default async function duplicateRequestPreventor(
+  req: Request,
+  res: Response,
+  next: NextFunction
+) {
+  const value = hashedBody(req);
+
+  if (queue.has(value)) return sendError(res, ERROR_MESSAGE, 429);
```
ChaituVR, line 15: Maybe we should use `409` error code?

ChaituVR (approved):
tAck

src/helpers/duplicateRequestPreventor.ts
wa0x6e: Or maybe `425` ?

PR #221, opened 2023-10-26
chore: stop logging 429 errors

wa0x6e:
Stop logging errors for rate limiting.

This is polluting the logs, and those requests are already tracked via the dashboard


ChaituVR (approved):
utAck

PR #222, opened 2023-10-27
refactor: extract redis to file

wa0x6e:
- Extract redis config to its own file, to enable reuse
- Enable redis in test suite

Waiting for https://github.com/snapshot-labs/actions/pull/16 merge to fix tests


Todmy (approved):
utACK

src/helpers/rateLimit.ts
```diff
@@ -34,10 +18,10 @@ export default rateLimit({
       429
     );
   },
-  store: client
+  store: redisClient
     ? new RedisStore({
-        sendCommand: (...args: string[]) => client.sendCommand(args),
-        prefix: 'snapshot-sequencer:'
+        sendCommand: (...args: string[]) => redisClient.sendCommand(args),
+        prefix: process.env.RATE_LIMIT_KEYS_PREFIX || 'snapshot-sequencer:'
```
ChaituVR, line 24: It is highly unlikely that we change this prefix, i feel it is better to leave it hard coded than adding a new env

ChaituVR: I think it can be the same prefix in test cases too

wa0x6e: This is to have a different prefix in test env

wa0x6e: By convention, test databases should always be separated from prod/local database, or it can lead to some edge cases later. It should be fine with using the same, but will not be consistent with the mysql database convention, which is already using separate

ChaituVR: Maybe we can make use of `process.env.NODE_ENV === 'test'` here

wa0x6e: And use different prefix, depending on node env ? Will be the same in the end, no? Different prefix for different env

ChaituVR: Yes yes, all I mean is removing `RATE_LIMIT_KEYS_PREFIX` from `.env`, it is growing with all optional params 😅

wa0x6e: It’s optional, and use only on test env, prod have a default

wa0x6e: Hard coding node_env check is not more elegant

ChaituVR (approved):
utAck

PR #225, opened 2023-10-28
fix: ensure snapshot value is in the network range

### LINKED ISSUE #224: fix: check that `snapshot` value is not lower than network first block

wa0x6e:

There are no check on the proposal's `snapshot` minimum value. People can submit very small block number not indexed by the nodes, resulting in error while computing voting power and scores.

## 💊 Fixes / Solution

Fix #224 

We should check that the snapshot value is in the range of the chosen network, by ensuring that `snapshot` is always equal or greater than the `start` property in networks.json

## 🚧 Changes

- Check that `snapshot` >= network's `start` block

## 🛠️ Tests

- Submit a proposal with a very small snapshot, it should fail with a `proposal snapshot must be after network start` error


ChaituVR (approved):
utAck

PR #226, opened 2023-10-30
fix: add a more aggressive rate limit for high errored users

wa0x6e:
From https://github.com/orgs/snapshot-labs/projects/12/views/7?pane=issue&itemId=43069387


yarn.lock
```diff
@@ -2760,9 +2760,9 @@ express-prom-bundle@^6.6.0:
     url-value-parser "^2.0.0"
 
 express-rate-limit@^6.9.0:
```
Todmy, line 2762: better to update package.json as well

Todmy (changes requested):

Todmy:
I have tried it and looks the case 
> Send some successful requests => you should be able to send 100 requests per minute
broken for me. After a short investigation, I see that the second rate limit you have added somehow overrides `windowMs` time for the first middleware.

wa0x6e:
> I have tried it and looks the case
> 
> > Send some successful requests => you should be able to send 100 requests per minute
> 
> broken for me. After a short investigation, I see that the second rate limit you have added somehow overrides `windowMs` time for the first middleware.

What do you mean by override? It resets the time?

wa0x6e:
The second, more aggressive rate limit will take precedence in case it is triggered.

First rate limit will work as usual, until there is 15 errored requests in a 15 seconds window, in which case, the user is locked completely until the 15 second window closes (every requests return 429, success or not)

Todmy:
> > I have tried it and looks the case
> > > Send some successful requests => you should be able to send 100 requests per minute
> > 
> > 
> > broken for me. After a short investigation, I see that the second rate limit you have added somehow overrides `windowMs` time for the first middleware.
> 
> What do you mean by override? It resets the time?

you can try it by yourself. With the current code, the first middleware doesn't work. 
I said `override` because I have inspect the keys stored in redis and found out that both keys are using 15s window

Todmy:
@wa0x6e ^^^

wa0x6e:
> @wa0x6e ^^^

Just tested, and confirming that the latest define rateLimit's `windowMs` is used. Swapping the order, and every limiter are using 60s instead.

Will see why

Todmy:
I can't test. Could you @ChaituVR  test take it?

wa0x6e:
> I can't test. Could you @ChaituVR test take it?

Will do, thank you

PR #228, opened 2023-10-31
Fix: Rename types.json to hashedTypes.json

ChaituVR:
More details here: https://github.com/snapshot-labs/snapshot.js/pull/927


wa0x6e (approved):
utAck

PR #230, opened 2023-11-01
Update dependabot.yml

samuveth:
This comment no longer applies as we are updating way more than just snapshot.js


ChaituVR (approved):
utAck

PR #232, opened 2023-11-03
feat: Handle unflag space and proposal

ChaituVR:
More context https://discord.com/channels/955773041898573854/1031838696846131200/1169949574786453505
New unflag is added in https://github.com/snapshot-labs/laser/pull/23

Unflagging a space and proposal if something is marked as spam by mistake


src/api.ts
```diff
@@ -72,9 +72,11 @@ router.post('/flag', verifyAuth, async (req, res) => {
   const { type, value, action } = req.body;
 
   try {
-    await flagEntity({ type, value, action });
+    const resp = await flagEntity({ type, value, action });
+    if (!resp.affectedRows) throw new Error('no rows affected');
```
wa0x6e, line 76: Throwing here will make the endpoint return a 500 error. 

Is it really what we want, or should we just use `return res.json({ success: false });` instead, and keep a 200 error code ?

This will let caller distinguish between a genuine server error, from a data error (given ID does not exist)

ChaituVR:
```suggestion
```

wa0x6e (approved):
tAck

PR #234, opened 2023-11-04
fix: rename `_at` suffix from Space timestamp properties

wa0x6e:
Ref to https://github.com/snapshot-labs/snapshot-hub/pull/740

Rename the Space's `created_at` and `updated_at` properties to just `created` and `updated`, for consistency


ChaituVR (approved):
utAck

PR #238, opened 2023-11-10
fix: prevent hibernated spaces from creating proposals

wa0x6e:
Related to pitch https://github.com/orgs/snapshot-labs/projects/12/views/7?pane=issue&itemId=43161232

Add a validation, and reject on proposal creation, when the space is hibernated

### DIFF
--- a/src/writer/proposal.ts
+++ b/src/writer/proposal.ts
@@ -64,6 +64,10 @@ export async function verify(body): Promise<any> {
     return Promise.reject('unknown space');
   }
 
+  if (space.hibernated) {
+    return Promise.reject('space hibernated');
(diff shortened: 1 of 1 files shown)


ChaituVR:
I feel like we should merge this only after having a way to un-hibernate (button on UI to retry as discussed) we may get many tickets

wa0x6e:
> I feel like we should merge this only after having a way to un-hibernate (button on UI to retry as discussed) we may get many tickets

Indeed, this is part 2 in the pitch implementation, and should be merged only when UI is also ready

ChaituVR (approved):
utAck

PR #244, opened 2023-11-17
fix: disable settings  validation on space creation

wa0x6e:
Do not validation space settings on proposal creation


ChaituVR (approved):
utAck

PR #254, opened 2023-11-25
fix: enable network validation in settings

wa0x6e:
Validate the network via schema validation

- Only allow mainnet networks on production
- Allow all networks on demo

### TODO 

- [x] Add tests
- [x] Waiting for https://github.com/snapshot-labs/snapshot.js/pull/945, to return correct error message (should be changed to `network not allowed`)


test/unit/writer/settings.test.ts
```diff
@@ -62,9 +62,11 @@ describe('writer/settings', () => {
         );
       });
 
-      // it('rejects if the network is on the wrong realm', async () => {
-      //   return expect(verify(editedInput({ network: '5' }))).rejects.toContain('wrong network');
-      // });
+      it('rejects if using testnet on production', async () => {
+        return expect(verify(editedInput({ network: '5' }))).rejects.toContain(
+          'must be a valid network used by snapshot'
```
wa0x6e, line 67: Need to be changed to `network not allowed` once https://github.com/snapshot-labs/snapshot.js/pull/945 and package updated

ChaituVR (approved):
tAck

PR #256, opened 2023-11-27
fix: disable proposal creation on misconfigured spaces

wa0x6e:
Following UI implementation https://github.com/snapshot-labs/snapshot/pull/4403

This PR disable proposals creation when the space is misconfigured.


src/writer/proposal.ts
```diff
@@ -36,6 +37,35 @@ export const getProposalsCount = async (space, author) => {
...
+
+  if (space?.deleted) return Promise.reject('space deleted, contact admin');
+
+  const schemaIsValid: any = snapshot.utils.validateSchema(snapshot.schemas.space, space, {
+    snapshotEnv: SNAPSHOT_ENV
+  });
+
+  if (schemaIsValid !== true) {
+    const firstErrorObject: any = Object.values(schemaIsValid)[0];
+    if (firstErrorObject.message === 'network not allowed') {
```
samuveth, line 57: If this this not true it should still `reject`, all space settings need to be valid

wa0x6e: All remaining errors from this validation are schema error (missing fields, wrong field type), which should never happen, since the space was fetched from the db directly via `getSpace()`.

All other settings (strategies, validation, etc...) are checked inside `validateSpaceSettings()`

ChaituVR: I think better to handle everything, instead of adding multiple if conditions in future (we may not remember that this if condition exist) ust in case if we add some other thing similar to networks.

samuveth (approved):
utACK

wa0x6e:
Tests fixed, most of the changed files are test files

src/writer/proposal.ts
```diff
@@ -36,6 +37,42 @@ export const getProposalsCount = async (space, author) => {
...
+async function validateSpace(originalSpace: any) {
+  const space = snapshot.utils.clone(originalSpace);
+
+  if (!space) {
+    return Promise.reject('unknown space');
+  }
+
+  if (space.hibernated) {
+    return Promise.reject('space hibernated');
+  }
```
ChaituVR, line 49: We should handle these inside `validateSpaceSettings`

src/writer/proposal.ts
```diff
@@ -36,6 +37,42 @@ export const getProposalsCount = async (space, author) => {
...
+
+  if (space?.deleted) return Promise.reject('space deleted, contact admin');
+
+  delete space.flagged;
+  delete space.verified;
+  delete space.id;
+
+  const schemaIsValid: any = snapshot.utils.validateSchema(snapshot.schemas.space, space, {
+    snapshotEnv: SNAPSHOT_ENV
+  });
```
ChaituVR, line 59: Better to handle these also in `validateSpaceSettings ` and reuse same function on both `settings` and `proposal`

ChaituVR (changes requested):

src/writer/proposal.ts
wa0x6e: Those check are proposal oriented, as those condition should not fail when updating settings

src/writer/settings.ts
ChaituVR, line 1: Somethings wrong i always get 
```
Wrong space format {"0":{"instancePath":"","keyword":"additionalProperties","message":"must NOT have additional properties","params":{"additionalProperty":"deleted"},"schemaPath":"#/additionalProperties"}}
```

ChaituVR (changes requested):

ChaituVR (approved):
tAck

Sorry for delay :)

PR #261, opened 2023-12-06
fix: handle missing uri when using poke

### LINKED ISSUE #260: FetchError: invalid json response body at https://cloudflare-ipfs.com/ipns/storage.snapshot.page/registry/0xe...
Sentry Issue: [SNAPSHOT-SEQUENCER-17](https://snapshot-labs.sentry.io/issues/4454216175/?referrer=github_integration)

```
FetchError: invalid json response body at https://cloudflare-ipfs.com/ipns/storage.snapshot.page/registry/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7/fabien.eth reason: Unexpected token i in JSON at position 2
  File "/workspace/build/src/helpers/ens.js", line 17, in getSpaceENS
    space = await snapshot_js_1.default.utils.getJSON(uri);
  File "/workspace/build/src/api.js", line 59, in <anonymous>
    const result = await (0, ens_1.getSpaceENS)(key);
```

wa0x6e:

When using poke on a ENS domain with an invalid `snapshot` TXT record, it fails with a JSON Parse error.

Sentry is capturing those errors, which are invalid user input, adding noises to errors logging

## 💊 Fixes / Solution

Fix #260 
Fix [SNAPSHOT-SEQUENCER-17](https://snapshot-labs.sentry.io/issues/4454216175/?referrer=github_integration)

Instead of logging the errors on Sentry, and returning `false` to user, we should return a more meaningfull error message to user, and stop logging those user errors on Sentry

## 🚧 Changes

- Add a try/catch on most error prone actions, and return the corresponding error message
- Add tests

## 🛠️ Tests

Try sending a GET spaces/:key/poke requests with different scenario

- Invalid ENS name
- ENS name without the `snapshot` TXT record
- ENS name with an invalid link in the TXT record
- ENS name with a TXT record linking to a not json file
- ENS name with a TXT record linking to a json file, but not following the space json format
- ENS with valid JSON file

On error, it should show an error message, instead of just `false` and should stop logging the error on Sentry


ChaituVR (approved):
tAck

PR #265, opened 2023-12-19
chore: Remove duplicate log

ChaituVR:
we already have a log at https://github.com/snapshot-labs/snapshot-sequencer/blob/remove-unused-logs/src/helpers/shutter.ts#L94 so this is just removing duplicate one


wa0x6e (approved):
utAck

PR #268, opened 2023-12-21
feat: New proposal limits for turbo spaces

ChaituVR:
Context https://discord.com/channels/955773041898573854/1187284357799481484/1187287152426504282

This PR will add new limits to turbo subscription, which is very similar to verified property 

How to test:
Change your space into turbo by changing it in DB

Example:
```SQL
update
  `spaces`
set
  `turbo` = '1'
where
  `id` = 'thanku.eth';
```

Change limits to something less
```JS
export const TURBO_SPACE_PROPOSAL_DAY_LIMIT = 1;
export const TURBO_SPACE_PROPOSAL_MONTH_LIMIT = 1;
```

While creating proposal on UI, you should proposal limit reached error


ChaituVR:
I think i need to add test cases also

wa0x6e (approved):
tAck

PR #275, opened 2024-01-10
feat: Change proposal body limit and strategies limit for turbo spaces

ChaituVR:
Change proposal body limit and strategies limit based on space type

### How to test:
- Use UI from https://github.com/snapshot-labs/snapshot/pull/4495/
- Add 10 strategies in space settings for turbo spaces
- Add 8 strategies in space settings for normal spaces
- Proposal body with 20000 characters for turbo spaces
- Proposal body with 10000 characters for normal spaces 
- Update proposal with different chars length

### DIFF
--- a/test/unit/writer/settings.test.ts
+++ b/test/unit/writer/settings.test.ts
@@ -9,6 +9,14 @@ function editedInput(payload = {}) {
   return { ...result, msg: JSON.stringify(result.msg) };
 }
 
+function randomStrategies(count = 1) {
+  return Array(count)
+    .fill(0)
+    .map(() => ({
+      name: `strategy-${Math.floor(Math.random() * 1000)}`
+    }));
+}
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const mockGetSpace = jest.fn((_): any => {
   return spacesGetSpaceFixtures;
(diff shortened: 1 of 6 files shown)


test/unit/writer/settings.test.ts
```diff
@@ -72,6 +80,27 @@ describe('writer/settings', () => {
...
+        mockGetSpace.mockResolvedValueOnce({ ...spacesGetSpaceFixtures, turbo: true });
+        return expect(
+          verify(
+            editedInput({
+              strategies: randomStrategies(12)
+            })
+          )
+        ).rejects.toContain('wrong space format');
+      });
     });
```
ChaituVR, line 104: Every time we want to increase number of strategies limit in snapshot.js, we have to change the limits here 🙈can i change it to pick from schema?

wa0x6e: Yes, was not aware it was exported

ChaituVR: or should be fine maybe?

wa0x6e: Dynamic will be better

wa0x6e (approved):
tAck

PR #278, opened 2024-01-15
fix: fix invalid scores array length after choices update

### LINKED ISSUE #276: Bug: Updated proposal's score and scores_by_strategy is wrong
If you look at values in these https://snapshot.org/#/thanku.eth/proposal/0xd99092aff1078a459bd99773371b8304438afa959cf817c2c46d5805022a7118
image

which is causing this error 
image

wa0x6e:

When updating a proposal with new choices, and when the `scores` and `scores_by_strategy` properties have been updated with empty values (eg. `[[0]]` instead of default `[]`), these fields are not updated, causing them to not have same number of element as the new choices element.

This cause errors on the UI side, which expect a `score` for each `choice`.

## 💊 Fixes / Solution

Fix #276 

## 🚧 Changes

- On proposal update, reset the `scores` and `scores_by_strategy` to an empty array. This reset does not cause side effect, as pending proposals does not have any scores yet.
- Also disable any `scores` and `scores_by_strategy` update when proposal is pending

### DIFF
--- a/src/scores.ts
+++ b/src/scores.ts
@@ -100,7 +100,7 @@ const pendingRequests = {};
 
 export async function updateProposalAndVotes(proposalId: string, force = false) {
   const proposal = await getProposal(proposalId);
-  if (!proposal) return false;
+  if (!proposal || proposal.state === 'pending') return false;
   if (proposal.scores_state === 'final') return true;
 
   if (!force && proposal.privacy === 'shutter' && proposal.state === 'closed') {
--- a/src/writer/update-proposal.ts
+++ b/src/writer/update-proposal.ts
@@ -67,11 +67,13 @@ export async function action(body, ipfs): Promise<void> {
     ipfs,
     updated,
     type: msg.payload.type,
-    plugins: plugins,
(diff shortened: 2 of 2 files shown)


src/scores.ts
```diff
@@ -100,7 +100,7 @@ const pendingRequests = {};
 
 export async function updateProposalAndVotes(proposalId: string, force = false) {
   const proposal = await getProposal(proposalId);
-  if (!proposal) return false;
+  if (!proposal || proposal.state === 'pending') return false;
```
ChaituVR, line 103: But `getProposal` doesn't return `state` property no?

ChaituVR (changes requested):

src/scores.ts
wa0x6e: It's used 3 lines below, at line 106, so I believe it's set

ChaituVR: Oh its set, i was looking at different `getProposal`

ChaituVR (approved):
tAck

PR #282, opened 2024-01-23
fix: remove dead moderation code

wa0x6e:

Some dead code in moderation.ts

## 💊 Fixes / Solution

Remove dead code

## 🚧 Changes

- Remove dead code in moderation for proposals body and title keywords, which are not used for moderation anymore
- Improve the moderation API links to fetch only the needed list

## 🛠️ Tests

- N/A


ChaituVR (approved):
utAck

better we should also remove flaggedProposalTitleKeywords, flaggedProposalBodyKeywords from sidekick and laser as well

PR #284, opened 2024-01-23
fix: auto flag proposal when the body contains flagged links

wa0x6e:

Proposals with flaggedLinks and proposals flagged manually are handled differently.

## 💊 Fixes / Solution

Auto-flag proposals containing flagged links, so the the only source of truth for flagging is the `flagged` column in the proposals table

## 🚧 Changes

- Load `flaggedLinks` data from moderation endpoint
- When creating and updating proposals, flag proposal when the body contains flagged links

## 🛠️ Tests

- create some proposals with body containing flagged links
- the proposal should be flagged, without manual moderation

### DIFF
--- a/src/helpers/moderation.ts
+++ b/src/helpers/moderation.ts
@@ -10,37 +10,48 @@ export let flaggedIps: Array<string> = [];
 export let flaggedAddresses: Array<string> = [];
 export let flaggedLinks: Array<string> = [];
 
-export async function loadModerationData(url = moderationURL): Promise<boolean> {
+export async function loadModerationData(
+  url = moderationURL
+): Promise<Record<string, string[]> | undefined> {
   try {
     const res = await fetchWithKeepAlive(url, { timeout: 5e3 });
     const body = await res.json();
 
     if (body.error) {
       capture(body);
-      return false;
+      return;
     }
 
-    flaggedIps = body.flaggedIps;
-    flaggedLinks = (body.flaggedLinks || []).filter((a: string) => a?.length > 0);
-    flaggedAddresses = (body.flaggedAddresses || []).map((a: string) => a.toLowerCase());
-
-    return true;
+    return {
+      flaggedIps: body.flaggedIps,
+      flaggedLinks: body.flaggedLinks,
+      flaggedAddresses: body.flaggedAddresses
+    };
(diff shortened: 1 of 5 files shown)


src/writer/proposal.ts
```diff
@@ -233,8 +233,10 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
     scores_total: 0,
     scores_updated: 0,
     votes: 0,
-    validation
+    validation,
+    flagged: containsFlaggedLinks(msg.payload.body) ? 1 : 0
```
ChaituVR, line 237: It should be the same as `update-proposal.ts` where it is `flagged: +containsFlaggedLinks(msg.proposal.body)` or we need to update `update-proposal.ts` like here 😅

src/helpers/moderation.ts
```diff
@@ -35,6 +37,13 @@ export default async function run() {
   run();
 }
 
+export function containsFlaggedLinks(body: string, links = flaggedLinks): boolean {
+  const normalizedLinks = links.filter(a => a?.length > 0);
```
ChaituVR, line 41: Better to move this out of this function, maybe to line `24`

ChaituVR (changes requested):

wa0x6e:
Moderation.ts has been refactored a bit, to accomodate tests

src/writer/update-proposal.ts
```diff
@@ -73,7 +74,8 @@ export async function action(body, ipfs): Promise<void> {
     discussion: msg.payload.discussion,
     choices: JSON.stringify(msg.payload.choices),
     scores: JSON.stringify([]),
-    scores_by_strategy: JSON.stringify([])
+    scores_by_strategy: JSON.stringify([]),
+    flagged: +containsFlaggedLinks(msg.proposal.body)
```
ChaituVR, line 78: but it is `msg.payload.body` in line 73 🙈

ChaituVR (changes requested):

src/writer/update-proposal.ts
wa0x6e: The danger of copilot autocompletion

ChaituVR (approved):
tAck

PR #292, opened 2024-02-24
feat: store proposal, vote and follower count on space

bonustrack:
I've created 3 new columns on the "spaces" table, `proposal_count`, `vote_count` and `follower_count`, this get automatically updated if some action happening on the space, this will allows us later to remove some heavy queries that we have currently on the hub.


wa0x6e:
Need to update the schema.sql on the hub with the new column, and reflect it to `https://github.com/snapshot-labs/snapshot-sequencer/blob/master/test/schema.sql`

src/writer/unfollow.ts
```diff
@@ -1,10 +1,19 @@
 import db from '../helpers/mysql';
 
-export async function verify(): Promise<any> {
+export async function verify(message): Promise<any> {
+  const query = `SELECT * FROM follows WHERE follower = ? AND space = ? LIMIT 1`;
```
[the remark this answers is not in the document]
bonustrack: How is that very expensive? We have indexes for follower and space.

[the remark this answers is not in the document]
wa0x6e: See https://github.com/snapshot-labs/snapshot-hub/issues/799 and https://snapshot-labs.sentry.io/performance/database/?end=2024-02-19T19%3A10%3A50&start=2024-02-19T09%3A49%3A05

src/writer/delete-proposal.ts
```diff
@@ -20,11 +20,14 @@ export async function verify(body): Promise<any> {
...
   const query = `
-  DELETE FROM proposals WHERE id = ? LIMIT 1;
-  DELETE FROM votes WHERE proposal = ?;
+    DELETE FROM proposals WHERE id = ? LIMIT 1;
+    DELETE FROM votes WHERE proposal = ?;
+    UPDATE spaces SET proposal_count = proposal_count - 1, vote_count = vote_count - ? WHERE id = ?;
   `;
-  await db.queryAsync(query, [id, id]);
+
+  await db.queryAsync(query, [id, id, proposal.votes, msg.space]);
```
ChaituVR, line 32: We are increasing `vote_count` on every new vote, but here `proposal.votes` is not guaranteed to be all votes, especially in case of override strategies where `proposal.votes` is updated once in a while. maybe better to get length from `votes` table

ChaituVR (changes requested):

src/writer/delete-proposal.ts
wa0x6e: This is counter is ambiguous. By using true votes count (number of table rows), it will unsync with the results in some case, where we could see 1 vote in the votes list, but 0 votes in the results.

This is an ongoing issue with all the proposal with failed score state.

wa0x6e: If we unsync the votes with results (which is the best solution for now), we should add a marker next to the votes on UI to let people know the vote is pending, so they know it’s not in the results yet

bonustrack: > We are increasing `vote_count` on every new vote, but here `proposal.votes` is not guaranteed to be same as all votes, especially in the case of override strategies where `proposal.votes` is updated once in a while.
> 
> Better to get length from the `votes` table

In which case vote count wouldn't be correct? Even if with override strategy someone may endup having 0 voting power, his vote would still be on the "votes" table. If we take votes length from the table it will be hard to update live, we would need to wait the end of the proposal to do the update which is not ideal.

ChaituVR: I mean we don't have to worry about results, if this count depends on results, then better to update this count when proposal is closed and not while inserting votes

ChaituVR: > if we take votes length from the table it will be hard to update live

I mean while deleting proposal, we should reduce based on votes count not with `proposal.votes`

ChaituVR: wyt? @bonustrack

wa0x6e: My take would be to keep 2 counters:

- `vote_count`: total number of votes for the proposals, regardless of status
- `(valid|executed|computed|resolved|whatever)_vote_count`: total number of votes with a valid final score (will exclude pending score from override strategies, as well as vote with invalid `score_state` see #280 )

bonustrack: Sound complicated, unless I'm missing something it should be rare case to have vote count unsync, and we can always resync / recompute the counter

wa0x6e: Everything that is not `final` is an unsynced counter:

Screenshot 2024-04-06 at 14 34 11

wa0x6e: > Sound complicated, unless I'm missing something it should be rare case to have vote count unsync, and we can always resync / recompute the counter

An example of unsynced counter: https://github.com/snapshot-labs/snapshot-hub/issues/579

wa0x6e: Refactored to always use total number of votes

wa0x6e (approved):
tAck, but need one more, as I made some edit

scripts/refresh_spaces_counters.ts
```diff
@@ -0,0 +1,35 @@
...
+    );
+
+    for (const stat of stats) {
+      await db.queryAsync(`UPDATE spaces SET vote_count = ?, proposal_count = ? WHERE id = ?`, [
+        stat.vote_count,
+        stat.proposal_count,
+        spaces[i].id
+      ]);
+      console.log(`${i} / ${spaces.length}`);
+    }
```
ChaituVR, line 23: We get only one value in `stats` no (because we get values of one space)?  why we need a `for` loop or `GROUP BY space` ?

I think it can be just
```suggestion
    const stats = await db.queryAsync(
      `SELECT COUNT(voter) as vote_count, COUNT(DISTINCT(proposal)) as proposal_count FROM votes WHERE space = ?`,
      [spaces[i].id]
    );
    const stat = stats[0];
    await db.queryAsync(`UPDATE spaces SET vote_count = ?, proposal_count = ? WHERE id = ?`, [
      stat.vote_count,
      stat.proposal_count,
      spaces[i].id
    ]);
    console.log(`${i} / ${spaces.length}`);
```

scripts/refresh_spaces_counters.ts
```diff
@@ -0,0 +1,35 @@
...
+// Usage: yarn ts-node scripts/refresh_spaces_counters.ts
+async function main() {
+  const spaces = await db.queryAsync(`SELECT id FROM spaces`);
+
+  console.log(`Found ${spaces.length} spaces`);
+
+  for (const i in spaces) {
+    const stats = await db.queryAsync(
+      `SELECT COUNT(voter) as vote_count, COUNT(DISTINCT(proposal)) as proposal_count FROM votes WHERE space = ? GROUP BY space`,
+      [spaces[i].id]
```
ChaituVR, line 13: I think this ignores proposals with `0` votes 

Maybe we can add a separate query to the `proposals` table

ChaituVR (changes requested):

wa0x6e:
@ChaituVR What's blocking here ?

scripts/refresh_spaces_counters.ts
```diff
@@ -0,0 +1,40 @@
...
+    console.log(`${i} / ${spaces.length}`);
+  }
+
+  await db.queryAsync(
+    'UPDATE spaces SET proposal_count = (SELECT count(id) from proposals WHERE space = spaces.id)'
+  );
+
+  await db.queryAsync(
+    'UPDATE spaces SET follower_count = (SELECT count(id) from follows WHERE space = spaces.id)'
+  );
```
ChaituVR, line 29: Not sure if these work correctly 🤔  @wa0x6e you tried?

wa0x6e: Yes, they should work

ChaituVR: Working :D i was confused

ChaituVR (approved):
tAck

PR #294, opened 2024-02-25
refactor: move get-quorum to scripts folder

ChaituVR:
We have two folders, so moving it to scripts folder to top level folder


wa0x6e (approved):
utAck

PR #315, opened 2024-03-19
fix: Store quorum type on proposal

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/issues/4600

- [x] Merge only after adding quorum_type to proposals table
```
ALTER TABLE proposals
ADD COLUMN quorum_type VARCHAR(24) DEFAULT '' AFTER quorum;
```

### How to test:
- Add `quorumType` on space settings
- On proposal creation, we should be able to store `quorumType` on proposals table

### DIFF
--- a/src/writer/proposal.ts
+++ b/src/writer/proposal.ts
@@ -245,6 +245,7 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
     start: parseInt(msg.payload.start || '0'),
     end: parseInt(msg.payload.end || '0'),
     quorum,
+    quorum_type: (quorum && spaceSettings.voting.quorumType) || '',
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

src/writer/proposal.ts
```diff
@@ -245,6 +245,7 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
     start: parseInt(msg.payload.start || '0'),
     end: parseInt(msg.payload.end || '0'),
     quorum,
+    quorum_type: (quorum && spaceSettings.voting.quorumType) || '',
```
ChaituVR, line 248:
```suggestion
    quorum_type: (quorum && spaceSettings.voting?.quorumType) || '',
```

Issue #309, opened 2024-03-11
Check for `snapshot` block availability on proposal creation


ChaituVR:
- While creating a proposal, Add a check if snapshot block is available or not
- This will check if user is passing a block that is not available on their network or RPC node we are using
- Also can remove this check
```js 
let currentBlockNum = 0;
  try {
    const provider = snapshot.utils.getProvider(space.network, { broviderUrl });
    currentBlockNum = parseInt(await provider.getBlockNumber());
  } catch {
    return Promise.reject('unable to fetch current block number');
  }

  if (msg.payload.snapshot > currentBlockNum)
    return Promise.reject('proposal snapshot must be in past');```

PR #319, opened 2024-03-25
Fix: On proposal creation check if snapshot block exist

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot-sequencer/issues/309

## Changes :
- Will check if `snapshot` block exists on the particular network
- Previously we were checking `snapshot` with the current block

## How to test :

- Go to your proposal creation page, for example https://snapshot.org/#/thanku.eth/create?snapshot=true
- Try creating proposal with future `snapshot`
- Untitled 2
- Before: it should throw error `proposal snapshot must be in past`
- Now: it should throw error `invalid snapshot block`

### DIFF
--- a/src/writer/proposal.ts
+++ b/src/writer/proposal.ts
@@ -172,19 +172,16 @@ export async function verify(body): Promise<any> {
     }
   }
 
-  let currentBlockNum = 0;
-  try {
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

PR #323, opened 2024-04-04
feat: add chainpatrol monitoring

### LINKED ISSUE #322: feat: scan for malicious link with ChainPatrol
We should use ChainPatrol API to scan for malicious link within proposals, similar as the BlockAid integration and using a config var for the API key.

ChaituVR:
Fixes #322 

- Also added a new folder monitoring to add these services
- Update blockaid to use API key from env
- Small refactoring

## How to test:
- Try to create proposals using spam links for example `daolido.org`
- Try to create proposals using genuine links


src/writer/proposal.ts
```diff
@@ -104,22 +104,19 @@ export async function verify(body): Promise<any> {
...
   try {
     const content = `
       ${msg.payload.name || ''}
       ${msg.payload.body || ''}
       ${msg.payload.discussion || ''}
     `;
-
     if (await isMalicious(content)) {
-      blockaidBlockedRequestsCount.inc({ space: space.id });
+      chainpatrolBlockedRequestsCount.inc({ space: space.id });
```
wa0x6e, line 114: Maybe move this one directly inside the chainpatrol helper

ChaituVR: no `space.id` in helper 🙈 is it worth to pass?

bonustrack (commented):
We don't need a folder "spamPrevention", also can you remove the harcoded API key from Blockaid, and use a config var instead?

wa0x6e:
Having a folder is good IMO, keep things structured

bonustrack:
If you want a folder call it "monitoring", this is not about spam

src/writer/proposal.ts
ChaituVR: there is no `space` name inside `proposal` data 😄

ChaituVR: will pass `space` as seperate param for now

wa0x6e (approved):
utAck

PR #331, opened 2024-04-23
fix: Update testnet space limit to be same as verified spaces

ChaituVR:
Full context: https://discord.com/channels/955773041898573854/1221106380300357804/1228248728964431932

Limit for normal space on testent will be the same as verified

### How to test:
- Try creating proposals on testnet, new limit will be `20` per day and `100` per month


wa0x6e (approved):
utAck

PR #333, opened 2024-04-25
fix: add params to restart leaderboard refresh script from arbitrary point

wa0x6e:
This PR refactor the leaderboard refresh script to accept an additional `pivot` argument, to restart the script from a specific `created` timestamp

Script now takes 2 optional arguments:

- `--space ID` : to filter by specific space ID (useful for migration)
- `--pivot TS` : to filter the spaces by `created`, and will target only spaces >= TS, useful to resume script at a specific point, when running on whole database.


ChaituVR (approved):
utAck

PR #334, opened 2024-04-25
fix: fix ER_LOCK_DEADLOCK error when inserting votes

wa0x6e:
remove the leaderboard query, debugging if it's the cause the sql `ER_LOCK_DEADLOCK` errors

### DIFF
--- a/src/writer/vote.ts
+++ b/src/writer/vote.ts
@@ -175,7 +175,7 @@ export async function action(body, ipfs, receipt, id, context): Promise<void> {
     );
   } else {
     // Store vote in dedicated table
-    await db.queryAsync('INSERT INTO votes SET ?;', [params, msg.space, voter]);
+    await db.queryAsync('INSERT INTO votes SET ?;', [params]);
   }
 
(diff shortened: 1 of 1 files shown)


src/writer/vote.ts
```diff
@@ -175,15 +175,7 @@ export async function action(body, ipfs, receipt, id, context): Promise<void> {
...
-    await db.queryAsync(
-      `
-        INSERT INTO votes SET ?;
-        INSERT INTO leaderboard (space, user, vote_count)
-          VALUES(?, ?, 1)
-          ON DUPLICATE KEY UPDATE vote_count = vote_count + 1
-      `,
-      [params, msg.space, voter]
-    );
+    await db.queryAsync('INSERT INTO votes SET ?;', [params, msg.space, voter]);
```
ChaituVR, line 178:
```suggestion
    await db.queryAsync('INSERT INTO votes SET ?;', [params]);
```

ChaituVR (approved):
utAck

PR #336, opened 2024-04-28
fix: prevent negative leaderboard counters

wa0x6e:
Fix https://snapshot-labs.sentry.io/issues/4542214149/events/044eed1cb2954dd5a772a29ac1b23eeb/?project=4505606761152512&query=is%3Aunresolved&referrer=previous-event&statsPeriod=90d&stream_index=5

Fix MySQL error where the `proposal_count` was set to a negative number.

As SQL transactions are not used, we can not guarantee that all counters are up-to-date


ChaituVR (approved):
utAck

PR #340, opened 2024-04-30
fix: Add testnet ecosystem spaces

ChaituVR:
Context: https://discord.com/channels/955773041898573854/1125390177888645231/1234850610612797451

We now need ecosystem limits for testnet spaces

### How to test:
- Change your local .env to use `NETWORK=testnet`
- Add a log here https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/proposal.ts#L184
- Try creating a proposal with testnet ecosystem space

### DIFF
--- a/src/helpers/limits.ts
+++ b/src/helpers/limits.ts
@@ -26,9 +26,7 @@ export const ECOSYSTEM_SPACES = [
   'arbitrumfoundation.eth'
 ];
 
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

wa0x6e:
Not really important IMHO, as long as it's not the other way

wa0x6e (approved):
utAck

PR #342, opened 2024-05-01
feat: add network support to follows

### LINKED ISSUE #341: feat: support follow for SX spaces
We need to support follow ability across our different protocol and have a single API to query those. 

To do so what I propose is to add a new column "network" (with index) on the table "follows", this column will use "s" for current follows and new follows for Snapshot space, "s-tn" for Snapshot testnet, "sep" for SX-EVM on Sepolia, "sn" for Starknet mainnet and so on. This param will allow us to query network specific follows. 

We may need to change length limit on space name to be able to use a long address (like Starknet contract address which are longer than on EVM). 

We may also need to change the "follow" action to support the "network" field.

We should also expose "network" within the Snapshot hub API, and have a filter so we can query for specific network.

wa0x6e:
Fix #341

This PR will add a new column `s` to the follows table, to be used to distinguish which network the space belongs to.

As of now, all past follows and new follows without `network` field will default to `s` (snapshot). v2 will bring more network choices.

This update will not affect anything on snapshot.org and SDK, current implementation should continue to work as usual.

Column length is 24 chars to have some margin, some network have long name (e.g `linea-testnet`)

## Changes

Add the new column to the database (may take a few minutes, depending on db size)

```sql
# For mainnet
ALTER TABLE follows ADD COLUMN network VARCHAR(24) NOT NULL DEFAULT 's' AFTER `space`;
# For testnet
ALTER TABLE follows ADD COLUMN network VARCHAR(24) NOT NULL DEFAULT 's-tn' AFTER `space`;
```

Update the existing primary key to include the new `network` column (took 50s, on my local DB with 5M rows, will lock the table while updating)

```sql
ALTER TABLE follows DROP PRIMARY KEY, ADD PRIMARY KEY (follower, space, network);
```

Add the missing indices on `space` and `network` column, because the composite index is not used when the wanted field is not the left-most index of the composite (see https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html)

```sql
ALTER TABLE `follows` ADD INDEX `space` (`space`);
ALTER TABLE `follows` ADD INDEX `network` (`network`);
```

## Tests 

- Start a local hub and sequencer (with this branch)
- Start a local instance of the frontend, using the local hub and sequencer
- JOIN and UNJOIN some spaces
- It should works (confirm on your local database)

## Todo 

- make follow readable by the hub: https://github.com/snapshot-labs/snapshot-hub/pull/847/files

### DIFF
--- a/src/writer/unfollow.ts
+++ b/src/writer/unfollow.ts
@@ -7,5 +7,5 @@ export async function verify(): Promise<any> {
 
 export async function action(message): Promise<void> {
   const query = 'DELETE FROM follows WHERE follower = ? AND space = ? AND network = ? LIMIT 1';
-  await db.queryAsync(query, [message.from, message.space, defaultNetwork]);
+  await db.queryAsync(query, [message.from, message.space, message.network || defaultNetwork]);
 }


src/writer/follow.ts
```diff
@@ -25,6 +25,7 @@ export async function action(message, ipfs, receipt, id): Promise<void> {
     ipfs,
     follower: message.from,
     space: message.space,
+    network: message.network || 's',
```
ChaituVR, line 28: Could we restrict `network` to be values we want?

ChaituVR (changes requested):

src/writer/follow.ts
ChaituVR: what all networks we want to add this?

wa0x6e: `network` is now validated

src/writer/follow.ts
```diff
@@ -25,6 +44,7 @@ export async function action(message, ipfs, receipt, id): Promise<void> {
...
```
ChaituVR, line 47: If it is `NETWORK` is `testnet`, we should default to `s-tn`

src/writer/unfollow.ts
```diff
@@ -5,6 +5,6 @@ export async function verify(): Promise<any> {
 }
 
 export async function action(message): Promise<void> {
-  const query = 'DELETE FROM follows WHERE follower = ? AND space = ? LIMIT 1';
-  await db.queryAsync(query, [message.from, message.space]);
+  const query = 'DELETE FROM follows WHERE follower = ? AND space = ? AND network = ? LIMIT 1';
+  await db.queryAsync(query, [message.from, message.space, message.network || 's']);
```
ChaituVR, line 9: Same here

test/schema.sql
```diff
@@ -106,9 +106,12 @@ CREATE TABLE follows (
   ipfs VARCHAR(64) NOT NULL,
   follower VARCHAR(64) NOT NULL,
   space VARCHAR(64) NOT NULL,
+  network VARCHAR(24) NOT NULL DEFAULT 's',
```
ChaituVR, line 109: Not sure how we can handle it in schema 🤔 maybe we should not default it to `s` but update values later

src/writer/follow.ts
```diff
@@ -1,6 +1,21 @@
...
+  'matic',
+  'arb1',
+  'oeth',
+  'gor',
+  'sep',
+  'linea-testnet',
+  'sn',
+  'sn-tn',
+  'sn-sep'
+];
```
ChaituVR, line 17: Maybe we can ignore `s-tn` ?

ChaituVR: Not sure if we want to have separate follow for every evm network, like someone with same address on `eth` `matic` `arb1` etc have to follow on each network?

ChaituVR (changes requested):

src/writer/follow.ts
wa0x6e: This is the network of the space, and not the voter

wa0x6e: Will make a different list for testnet and mainnet

test/schema.sql
wa0x6e: Leaving the field blank, then updating it later may result in duplicate field, as we need to update the primary index also, to include the new `network` field.

We could leave the schema as-is, and use the query in the PR description to set a different default value for different env.

src/writer/unfollow.ts
```diff
@@ -1,10 +1,11 @@
...
 
 export async function verify(): Promise<any> {
   return true;
 }
 
 export async function action(message): Promise<void> {
-  const query = 'DELETE FROM follows WHERE follower = ? AND space = ? LIMIT 1';
-  await db.queryAsync(query, [message.from, message.space]);
+  const query = 'DELETE FROM follows WHERE follower = ? AND space = ? AND network = ? LIMIT 1';
+  await db.queryAsync(query, [message.from, message.space, defaultNetwork]);
```
ChaituVR, line 10: why only `defaultNetwork`? users won't be able to unfollow on other networks?

ChaituVR (changes requested):

src/writer/unfollow.ts
ChaituVR: i think it should be `message.network || defaultNetwork,` similar to follow action

wa0x6e: Correct

ChaituVR (approved):
tAck

PR #343, opened 2024-05-02
feat: add `last_vote` to leaderboard

wa0x6e:
Add `last_vote` column to leaderboard

## Changes 

Run the following queries to add the new column and index

```sql
ALTER TABLE leaderboard ADD COLUMN last_vote BIGINT;
ALTER TABLE leaderboard ADD INDEX last_vote (`last_vote`);
```

## Test 

- Run the `refresh_leaderboard_counters` script
- It should update the `last_vote` field with the correct timestamp 
- Boot a local instance of the UI, connected to this branch
- Vote on a proposal
- The leaderboard should be updated with the correct timestamp (confirm in the DB directly)

### DIFF
--- a/src/helpers/actions.ts
+++ b/src/helpers/actions.ts
@@ -76,13 +76,13 @@ export function refreshVotesCount(spaces: string[]) {
     `
       INSERT INTO leaderboard (vote_count, user, space)
         (SELECT * FROM (
-          SELECT COUNT(votes.id) AS vote_count, voter, space
+          SELECT COUNT(votes.id) AS vote_count, MAX(created) as last_vote, voter, space
           FROM votes
           JOIN spaces ON BINARY spaces.id = BINARY votes.space
           WHERE spaces.deleted = 0 AND space IN (?)
           GROUP BY voter, space
         ) AS t)
-      ON DUPLICATE KEY UPDATE vote_count = t.vote_count
+      ON DUPLICATE KEY UPDATE vote_count = t.vote_count, last_vote = t.last_vote
     `,
     spaces
   );
--- a/src/helpers/actions.ts
+++ b/src/helpers/actions.ts
@@ -74,9 +74,9 @@ export function refreshProposalsCount(spaces?: string[]) {
 export function refreshVotesCount(spaces: string[]) {
   return db.queryAsync(
     `
-      INSERT INTO leaderboard (vote_count, user, space)
+      INSERT INTO leaderboard (vote_count, last_vote, user, space)
         (SELECT * FROM (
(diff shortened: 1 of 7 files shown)


src/helpers/actions.ts
```diff
@@ -76,13 +76,13 @@ export function refreshVotesCount(spaces: string[]) {
     `
       INSERT INTO leaderboard (vote_count, user, space)
         (SELECT * FROM (
-          SELECT COUNT(votes.id) AS vote_count, voter, space
+          SELECT COUNT(votes.id) AS vote_count, MAX(created) as last_vote, voter, space
```
ChaituVR, line 79: while running the script i get this error 

```bash
OperationalError: ER_NON_UNIQ_ERROR: Column 'created' in field list is ambiguous
    at Query.Sequence._packetToError (/snapshot-labs/snapshot-sequencer/
```

ChaituVR (changes requested):

src/helpers/actions.ts
ChaituVR: If i change it to `votes.created` then i get `OperationalError: ER_WRONG_VALUE_COUNT_ON_ROW: Column count doesn't match value count at row 1`

ChaituVR: what will be the value of last_vote if there are no votes for a address

wa0x6e: > what will be the value of last_vote if there are no votes for a address

NULL

ChaituVR: Any idea how to solve the above? :(

wa0x6e: SQL error fixed

src/writer/vote.ts
```diff
@@ -178,11 +178,11 @@ export async function action(body, ipfs, receipt, id, context): Promise<void> {
     await db.queryAsync(
       `
         INSERT INTO votes SET ?;
-        INSERT INTO leaderboard (space, user, vote_count)
-          VALUES(?, ?, 1)
-          ON DUPLICATE KEY UPDATE vote_count = vote_count + 1
+        INSERT INTO leaderboard (space, user, vote_count, last_vote)
+          VALUES(?, ?, 1, ?)
+          ON DUPLICATE KEY UPDATE vote_count = vote_count + 1, last_vote = ?
```
ChaituVR, line 183: Maybe we should also update leaderboard while updating existing vote? if I update choice it still use timestamp of first vote

wa0x6e: Yes, I forgot about the update vote scenario

wa0x6e: Added

ChaituVR (approved):
tAck

PR #344, opened 2024-05-02
refactor: improve refresh leaderboard script

wa0x6e:
This PR improves the refresh leaderboard script

## Changes

- Instead of batching the requests by spaces, it is now batched by voter, resulting in much smaller query, but with whole script taking longer due to huge number of voters
- the `pivot` arg is now about the votes created date, and not space
- Split the refresh proposals and votes into 2 actions

Refresh the proposals count with 

```
yarn ts-node scripts/refresh_leaderboard_counters.ts --type proposal
```

Refresh the votes count with 

```
yarn ts-node scripts/refresh_leaderboard_counters.ts --type vote
```

Both can be filtered by `--space` and vote have additional `pivot` (vote date)

### DIFF
--- a/scripts/refresh_leaderboard_counters.ts
+++ b/scripts/refresh_leaderboard_counters.ts
@@ -1,50 +1,60 @@
 import 'dotenv/config';
-import { refreshProposalsCount, refreshVotesCount } from '../src/helpers/actions';
 import db from '../src/helpers/mysql';
+import { refreshProposalsCount, refreshVotesCount } from '../src/helpers/actions';
 
 // Usage: yarn ts-node scripts/refresh_leaderboard_counters.ts --space [OPTIONAL-SPACE-ID] --pivot TIMESTAMP
 async function main() {
-  const filters: string[] = [];
+  let spacesFilter = '';
+  let usersFilter = '';
 
   process.argv.forEach((arg, index) => {
     if (arg === '--space') {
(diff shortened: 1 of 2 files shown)


scripts/refresh_leaderboard_counters.ts
```diff
@@ -1,50 +1,60 @@
...
+      console.log('Filtered by users.created >=', process.argv[index + 1]);
+      usersFilter = `WHERE created >= ${process.argv[index + 1].trim()}`;
     }
   });
+  console.log('');
 
-  const filtersQuery = filters.length ? `WHERE ${filters.join(' AND ')}` : '';
-  const query = `SELECT id, name, created FROM spaces ${filtersQuery} ORDER BY created ASC`;
+  const users: { id: string; created: number }[] = await db.queryAsync(
+    `SELECT id, created FROM users ${usersFilter} ORDER BY created ASC`
```
ChaituVR, line 26: But not every voter/proposal creator is on `users`

ChaituVR (changes requested):

scripts/refresh_leaderboard_counters.ts
wa0x6e: Ouch, was not aware of this

wa0x6e: Script updated to loop by voter instead

ChaituVR (approved):
tAck

PR #347, opened 2024-05-06
fix: enable alias voting and propose globally

wa0x6e:


ChaituVR (changes requested):
Initial plan was to allow alias voting if space allow it, but not sure what is the plan now, @bonustrack Could you confirm that we can allow everyone to vote with alias

ChaituVR (commented):
Sounds a bit dangerous because users don't have any option to delete their alias 
just in case if user login using a unknown device

Maybe before this could we add a way for users to delete their alias and have an expiry for alias? so that old alias won't be used to cast votes

wa0x6e:
> Sounds a bit dangerous because users don't have any option to delete their alias just in case if user login using a unknown device
> 
> Maybe before this could we add a way for users to delete their alias and have an expiry for alias? so that old alias won't be used to cast votes

Yes, ideally, we should have a section in user profile to list all alias, and ideally, have a way to customize which actions can be used with alias

src/ingestor.ts
```diff
@@ -58,32 +58,30 @@ export default async function ingestor(req) {
...
+      'unsubscribe',
+      'profile',
       'vote',
       'vote-array',
       'vote-string',
       'proposal',
+      'update-proposal',
       'delete-proposal',
       'statement'
     ];
```
ChaituVR, line 82: what action types are not this list except `settings` and `alias`? maybe we can remove this list and check for `settings` and `alias` types only?

ChaituVR:
waiting for #353

ChaituVR (approved):
utAck

PR #348, opened 2024-05-07
fix: Flagged link check to handle special chars

ChaituVR:
### Summary
- [ Handle special chars ](https://discord.com/channels/955773041898573854/1221106380300357804/1237318479829405716) eg: `clcula` is matching if flaggedLinks contains `clc.la`
- Detect even if the protocol is either `http` or `https` or no protocol at all
- Detect if the link contains a subdomain, for example using `www`
- Detect if the path is anything else than flagged one
- Detect if the link is case-insensitive

### How to test
- Run `yarn test:integration` which contains all the new cases


wa0x6e (approved):
utAck

PR #349, opened 2024-05-08
fix: remove deprecated goerli network from follows network

wa0x6e:
Following https://github.com/snapshot-labs/sx-monorepo/pull/327 

Remove goerli from supported follows network


ChaituVR (approved):
utAck

PR #351, opened 2024-05-11
feat: validate onchain space on follows action

wa0x6e:
This PR:

- moves the space validation for follow/unfollow actions to the writer (for custom validation)
- connect to sx api to validate space from sx

### How to test

- Boot a local instance of sx-monorepo, using this PR branch
- Follow and unfollow some spaces, it should work
- Boot a local instance of snapshot, using this PR branch 
- Follow and unfollow some spaces, it should work

### DIFF
--- a/src/helpers/actions.ts
+++ b/src/helpers/actions.ts
@@ -1,6 +1,7 @@
 import snapshot from '@snapshot-labs/snapshot.js';
 import db from './mysql';
 import { jsonParse } from './utils';
+import { defaultNetwork } from '../writer/follow';
 
 export async function addOrUpdateSpace(space: string, settings: any) {
   if (!settings?.name) return false;
@@ -38,7 +39,16 @@ export async function getProposal(space, id) {
   return proposal;
 }
 
-export async function getSpace(id: string, includeDeleted = false) {
+export async function getSpace(id: string, includeDeleted = false, network = defaultNetwork) {
+  if (network !== defaultNetwork) {
+    const spaceExist = await sxSpaceExists(id);
+    if (!spaceExist) return false;
+
+    return {
+      network: 1
+    };
+  }
+
   const query = `SELECT settings, deleted, flagged, verified, turbo, hibernated FROM spaces WHERE id = ? AND deleted in (?) LIMIT 1`;
   const spaces = await db.queryAsync(query, [id, includeDeleted ? [0, 1] : [0]]);
 
--- a/src/helpers/actions.ts
+++ b/src/helpers/actions.ts
@@ -1,3 +1,4 @@
+import snapshot from '@snapshot-labs/snapshot.js';
 import db from './mysql';
 import { jsonParse } from './utils';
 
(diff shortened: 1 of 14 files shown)


src/ingestor.ts
```diff
@@ -62,10 +62,14 @@ export default async function ingestor(req) {
...
-      if (!space) return Promise.reject('unknown space');
-      network = space.network;
-      if (space.voting?.aliased) aliased = true;
+
+      // Disable space checking for follow/unfollow, checking delegated to the writer
+      if (!['follow', 'unfollow'].includes(type)) {
+        const space = await getSpace(message.space);
+        if (!space) return Promise.reject('unknown space');
+        network = space.network;
+        if (space.voting?.aliased) aliased = true;
```
ChaituVR, line 71: I think this line should be outside `if` if not, aliased will `false` for `follow` and `unfollow`

src/ingestor.ts
```diff
@@ -62,10 +62,14 @@ export default async function ingestor(req) {
...
-      const space = await getSpace(message.space);
-      if (!space) return Promise.reject('unknown space');
-      network = space.network;
-      if (space.voting?.aliased) aliased = true;
+
+      // Disable space checking for follow/unfollow, checking delegated to the writer
+      if (!['follow', 'unfollow'].includes(type)) {
+        const space = await getSpace(message.space);
+        if (!space) return Promise.reject('unknown space');
+        network = space.network;
```
ChaituVR, line 70: maybe this line also?

ChaituVR (changes requested):

src/ingestor.ts
wa0x6e: But outside the if, `space` is undefined, so aliased will always be false anyway

src/ingestor.ts
```diff
@@ -62,10 +62,14 @@ export default async function ingestor(req) {
     let aliased = false;
     if (!['settings', 'alias', 'profile'].includes(type)) {
       if (!message.space) return Promise.reject('unknown space');
-      const space = await getSpace(message.space);
-      if (!space) return Promise.reject('unknown space');
-      network = space.network;
-      if (space.voting?.aliased) aliased = true;
+
+      // Disable space checking for follow/unfollow, checking delegated to the writer
+      if (!['follow', 'unfollow'].includes(type)) {
```
ChaituVR, line 67: I think we can include `follow` and `unfollow` in line `63` no?

wa0x6e: will lose the `message.space` params existence checking

wa0x6e:
PR refactor to include the SX spaces validation

ChaituVR (changes requested):

src/writer/follow.ts
```diff
@@ -27,6 +28,14 @@ export async function verify(message): Promise<any> {
...
   }
 
+  if (!message.network || message.network === defaultNetwork) {
+    const space = await getSpace(message.space);
+    if (!space) return Promise.reject('unknown space');
+  } else {
+    const spaceExist = await sxSpaceExists(message.space);
+    if (!spaceExist) return Promise.reject('unknown space');
+  }
+
```
ChaituVR, line 38: Same here

src/helpers/actions.ts
```diff
@@ -37,7 +39,16 @@ export async function getProposal(space, id) {
...
 }
 
-export async function getSpace(id: string, includeDeleted = false) {
+export async function getSpace(id: string, includeDeleted = false, network = defaultNetwork) {
+  if (network !== defaultNetwork) {
+    const spaceExist = await sxSpaceExists(id);
+    if (!spaceExist) return false;
+
+    return {
+      network: 1
```
ChaituVR, line 48: feel like this may trouble us in the future. 😢

wa0x6e: We could pass `0` instead

ChaituVR: hmm if it works then yea 😄

ChaituVR (approved):
utAck

PR #353, opened 2024-05-14
fix: Only use aliases that were created within the last 30 days

ChaituVR:
- Use alias that were created  within last 30 days 

How to test:
- Try following a space with a alias that was created 30 days ago

TODO:
- [x] Handle from UI https://github.com/snapshot-labs/snapshot/pull/4721


wa0x6e (approved):
utAck

PR #354, opened 2024-05-14
fix: Add script to initialize leaderboard counters

wa0x6e:

The current refresh_leaderboard_counter script is too slow to refresh counters for the whole database. It will be kept and still used for task such as space deletion and migration, but another script will be necessary for the counter initialization

## 💊 Fixes / Solution

Create another script to initialize leaderboard votes counter

## 🚧 Changes

- Get voters per day, and one day at a time until pivot === today
- Do not use `JOIN` query, for faster query
- Process one user/space at a time

There's still a `group by` in the query, that can be moved to nodejs in case the query is not performant enough

## 🛠️ Tests

- Run `yarn ts-node scripts/init_leaderboard_counters.ts`
- It should populate the `vote_count` and `last_vote` in the leaderboard tables

### DIFF
--- a/scripts/init_leaderboard_counters.ts
+++ b/scripts/init_leaderboard_counters.ts
@@ -1,0 +1,92 @@
+import 'dotenv/config';
+import db from '../src/helpers/mysql';
+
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

PR #356, opened 2024-05-18
fix: use whitelist to validate space network

wa0x6e:
This PR uses a whitelist to validate the `network` argument, and will check for sx space only if it's a valid SX network, compared to before, where it would consider all non-default-network as SX network.

This PR will fix an issue where people are still trying to pass the `network` params when creating a proposal


ChaituVR (approved):
utAck

PR #358, opened 2024-05-20
fix: Disable leaderboard updates

ChaituVR:
Context: https://discord.com/channels/955773041898573854/1031838169282392144/1242062206883856414

PR #360, opened 2024-05-21
fix: Add back leaderboard update queries

ChaituVR:
This is just adding back https://github.com/snapshot-labs/snapshot-sequencer/pull/358/files and update delete proposal to reduce count of proposal voters instead of refresh

### DIFF
--- a/src/writer/delete-proposal.ts
+++ b/src/writer/delete-proposal.ts
@@ -20,13 +20,23 @@ export async function verify(body): Promise<any> {
 
 export async function action(body): Promise<void> {
   const msg = jsonParse(body.msg);
+  const proposal = await getProposal(msg.space, msg.payload.proposal);
+  const voters = await db.queryAsync(`SELECT voter FROM votes WHERE proposal = ?`, [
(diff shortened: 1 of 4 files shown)


src/writer/delete-proposal.ts
```diff
@@ -20,13 +20,23 @@ export async function verify(body): Promise<any> {
...
     DELETE FROM votes WHERE proposal = ?;
+    UPDATE leaderboard
+      SET proposal_count = GREATEST(proposal_count - 1, 0)
+      WHERE user = ? AND space = ?
+      LIMIT 1;
+    UPDATE leaderboard SET vote_count = GREATEST(vote_count - 1, 0)
+      WHERE user IN (?) AND space = ?;
   `,
-    [id, id]
+    [id, id, proposal.author, msg.space, voters.map(voter => voter.voter), msg.space]
```
wa0x6e, line 40: Will this still work if voters is empty (`[]`) ?

wa0x6e (approved):
utAck

PR #362, opened 2024-05-21
fix: fix init script not processing all days after pivot

wa0x6e:
Fix init leaderboard script not processing all days after pivot correctly


ChaituVR (approved):
utAck

PR #365, opened 2024-05-22
fix: ignore 504 errors from score-api

wa0x6e:
Following this https://github.com/snapshot-labs/snapshot-sequencer/pull/359

Ignore 504 errors from missed validation


ChaituVR (approved):
utAck

PR #366, opened 2024-05-23
fix: improve leaderboard refresh script

wa0x6e:
This PR improve the refresh leaderboard script speed

- Remove the `BINARY` casting
- Process voters day by day, but a voter is only processed once (will refresh all time counters, and not just for the day)
- Added a --end arg, so we could run tell the script to stop at a given timestamp
- Rename `--pivot` to `--start`

On my local intance, this script is getting 350 requests per seconds. We have ~6M unique voters, and should thus take ~4hours to refresh the whole table.

As per my tests, running more script in parallel does not decrease the RPS, so we could start multiple instance of the script, with different `--end` and `--start` (such as one script per year), and may drop the total running time to ~1hour


ChaituVR (approved):
utAck

PR #369, opened 2024-05-27
fix: use correct SX API for different network

### LINKED ISSUE #355: fix: use correct sx api endpoint
We're currently using thegraph api for sx, due to https://api-1.snapshotx.xyz/ not returning correct results, because of https://github.com/checkpoint-labs/checkpoint/issues/296

Once it's fixed, we should move back to the correct API url, to enable also starknet spaces

wa0x6e:

The SX API we're currently using to check and validate onchain spaces on Follow actions is only for a single type of network

## 💊 Fixes / Solution

Fix #355 

Use correct SX API, for each different network types

## 🚧 Changes

- Use a list of different SX API, for each network, like on stamp https://github.com/snapshot-labs/stamp/blob/master/src/resolvers/space-sx.ts

## 🛠️ Tests

- Run `yarn test`

### DIFF
--- a/src/helpers/actions.ts
+++ b/src/helpers/actions.ts
@@ -41,7 +41,7 @@ export async function getProposal(space, id) {
 
 export async function getSpace(id: string, includeDeleted = false, network = defaultNetwork) {
   if (NETWORK_WHITELIST.includes(network) && network !== defaultNetwork) {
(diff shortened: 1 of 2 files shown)


ChaituVR (changes requested):
utAck

ChaituVR (approved):
utAck

PR #374, opened 2024-06-03
fix: enforce lowercase space ID

wa0x6e:
Fixes https://github.com/snapshot-labs/snapshot-hub/issues/448

This PR enforce that:

- Space ID is always lower case on space creation
- Checking for space existence is case insensitive, so `a.eth` and `A.eth` are equal (this is already the case with the current table collation, will be more useful once we moved to the new case sensitive collation)


ChaituVR (approved):
utAck

PR #376, opened 2024-06-04
fix: update sql schema for new space length in follows table

wa0x6e:
Schema update to match the new column length for space in follows table


ChaituVR (approved):
utAck

PR #382, opened 2024-06-11
feat: clear stamp cache on data update

wa0x6e:
Related to https://discord.com/channels/955773041898573854/1031838169282392144/1249097030094291066

This PR will call stamp service and clear the relevant cache on avatar update

Cache will be cleared when updating:

- user's avatar
- spaces' avatar

### DIFF
--- a/src/writer/profile.ts
+++ b/src/writer/profile.ts
@@ -1,6 +1,6 @@
 import snapshot from '@snapshot-labs/snapshot.js';
 import db from '../helpers/mysql';
-import { jsonParse } from '../helpers/utils';
+import { jsonParse, clearStampCache } from '../helpers/utils';
 import log from '../helpers/log';
 
 export async function verify(body): Promise<any> {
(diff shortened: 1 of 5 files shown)


src/writer/profile.ts
```diff
@@ -25,4 +29,6 @@ export async function action(message, ipfs): Promise<void> {
   };
 
   await db.queryAsync('REPLACE INTO users SET ?', params);
+
+  if (profile.avatar !== existingProfile.avatar) clearStampCache('avatar', message.from);
```
ChaituVR, line 33: Maybe better to wait until we finish the fetch, else user may see old avatar for some time,

ChaituVR (changes requested):

src/writer/profile.ts
wa0x6e: How could we achieve that ? What do you mean by finish the fetch ?

This is the method used by the UI (directly trigger this request after profile update), and I did notice that sometimes, the cache is not clearing correctly.

ChaituVR: I mean to wait until the cache is cleared.

ChaituVR (approved):
tAck

PR #383, opened 2024-06-16
fix: fix missing value for updated column

wa0x6e:
When trying to create a new statement, the query is failing with `ER_NO_DEFAULT_FOR_FIELD ` error, due to the `updated` field not having a default value

This PR set a default value when inserting a new statement


ChaituVR (approved):
utAck

PR #384, opened 2024-06-17
fix: Increase statements id length

ChaituVR:
Change the length of the `id` field in statements to be the same as other tables


wa0x6e (approved):
utAck

PR #389, opened 2024-06-18
fix: fix new statement properties being ignored

wa0x6e:
Fix new Statement properties missing from the payload


ChaituVR (approved):
utAck

PR #396, opened 2024-06-29
fix: Change length of address fields

ChaituVR:
Same as https://github.com/snapshot-labs/snapshot-hub/pull/877


test/schema.sql
```diff
@@ -76,7 +76,7 @@ CREATE TABLE proposals (
 CREATE TABLE votes (
   id VARCHAR(66) NOT NULL,
   ipfs VARCHAR(64) NOT NULL,
-  voter VARCHAR(64) NOT NULL,
+  voter VARCHAR(100) NOT NULL,
   created INT(11) NOT NULL,
   space VARCHAR(64) NOT NULL,
```
wa0x6e, line 81: `space` is also 100 on this votes table

wa0x6e (changes requested):

test/schema.sql
ChaituVR, line 81:
```suggestion
  space VARCHAR(100) NOT NULL,
```

wa0x6e (approved):
utAck

PR #401, opened 2024-07-05
fix: delete follows and subscriptions when deleting a space

wa0x6e:


ChaituVR:
Why delete follows and subscriptions 🤔 we should never delete anything ideally.

wa0x6e:
This is edge case, but if I follow a space, and it gets deleted, it will stay in my follows, and count toward my 25 spaces quota

ChaituVR:
True that, agree. but lets ask @bonustrack if he agree

ChaituVR:
or maybe we can ignore deleted spaces while checking for limit?

wa0x6e:
See https://github.com/snapshot-labs/snapshot-sequencer/pull/486

PR #406, opened 2024-07-12
fix: update refresh spaces counters script

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot-sequencer/pull/406

Context: DIscussed on discord


wa0x6e:
The new approach was to avoid running one query per space, since it's the `SELECT` part that was slow. 

Doing a `SELECT FROM leaderboard` for a single space was just a little faster than doing for thousands of space.

Approach will be:

- Send a single `SELECT GROUP BY space` to select all vote_count for a lot of space at once
- `UPDATE space` with the results from above query

Since the first query (`SELECT`) may be slow for big spaces, we can do 2 batches, one for verified spaces, and another one for non-verified spaces.

The SELECT query will look like

```
select SUM(leaderboard.vote_count), space from leaderboard JOIN spaces ON spaces.id = space   where spaces.verified = 1 GROUP by space 
```

PR #408, opened 2024-07-18
fix: revert .env.test to mainnet

ChaituVR:


wa0x6e (approved):
utAck

wa0x6e:
Maybe also revert the changes inside ingestor

https://github.com/snapshot-labs/snapshot-sequencer/blob/bda8b9d735cf8e0605dc73c2e1e4d1fa673508db/src/ingestor.ts#L29

To have everything other than `mainnet` to be `testnet`, like on the hub

ChaituVR:
i think we can leave it like this, we are checking testnet everywhere

PR #417, opened 2024-07-28
fix: validate space ID

wa0x6e:

We currently don't validate space ID before querying and inserting space id.

When querying from the database, an ID with special hidden character (`abc[][].eth`) will be auto-cleaned, and the ID will be stripped of all invalid characters (`abc.eth`), returning incorrect results.

Same when inserting, an ID with hidden special characters will be inserted with its cleaned version.

Currently, wrong insertion are prevented by the `getSpaceController()` on the SDK side, which is failing with an uncaught error.

## 💊 Fixes / Solution

This PR will add space ID validation on sequencer side also, rejecting all non-valid space ID by using ENS normalize function.

## 🚧 Changes

- Validate space ID before querying spaces from database
- Validate space ID when saving space settings

## 🛠️ Tests

- try sending the payload in https://snapshot-labs.sentry.io/issues/5639825201/?query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=15 (need to disable timestamp check on ingestor)
- Before fix, it throws an uncaught error, and request return `server_error`
- After fix, request should return `Invalid space ID` error

### DIFF
--- a/src/writer/settings.ts
+++ b/src/writer/settings.ts
@@ -1,4 +1,3 @@
-import { ensNormalize } from '@ethersproject/hash';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
 import isEqual from 'lodash/isEqual';
@@ -72,12 +71,6 @@ export async function verify(body): Promise<any> {
     });
   } catch (e) {
     return Promise.reject(e);
-  }
-
-  try {
-    if (ensNormalize(msg.space) !== msg.space.toLowerCase()) throw new Error('');
-  } catch (e) {
-    return Promise.reject('Invalid space id');
   }
 
   const controller = await snapshot.utils.getSpaceController(msg.space, DEFAULT_NETWORK, {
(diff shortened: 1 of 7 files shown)


ChaituVR (dismissed):
utAck

src/writer/settings.ts
```diff
@@ -73,6 +74,12 @@ export async function verify(body): Promise<any> {
     return Promise.reject(e);
   }
 
+  try {
+    if (ensNormalize(msg.space) !== msg.space.toLowerCase()) throw new Error('');
+  } catch (e) {
+    return Promise.reject('Invalid space id');
+  }
```
ChaituVR, line 81: I think better to do it at ingetor level, else we may have to do this on all actions?

wa0x6e:
GetSpace is already checking that

ChaituVR:
> GetSpace is already checking that

hmm I see, it is doing on all actions except `alias` and `profile`, maybe we can do single check on ingestor instead of two in `getSpace` and `settings`

it can be before line 82, 
```ts
  if (message.space && ensNormalize(message.space) !== message.space.toLowerCase()) return Promise.reject('invalid space id')
```

ChaituVR (approved):
utAck

PR #427, opened 2024-08-24
fix: validate space ID (bis)

wa0x6e:
Reopening same PR as #417 

But with an additional fix to validate space ID only for offchain spaces

### Test

- See #417 
- With this PR fix, we check space ID only if it's an ID sent by v1 (without network prefix), or if it's an offchain ID sent by v2 (with `s:` prefix)

### DIFF
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -82,7 +82,7 @@ export default async function ingestor(req) {
 
     try {
       if (
-        (message.space.startWith('s:') || !message.space.includes(':')) &&
+        (message.space.startsWith('s:') || !message.space.includes(':')) &&
         ensNormalize(message.space) !== message.space.toLowerCase()
       )
         throw new Error('');
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -82,6 +82,7 @@ export default async function ingestor(req) {
 
     try {
       if (
+        message.space &&
         (message.space.startsWith('s:') || !message.space.includes(':')) &&
         ensNormalize(message.space) !== message.space.toLowerCase()
       )
--- a/src/ingestor.ts
+++ b/src/ingestor.ts
@@ -81,7 +81,11 @@ export default async function ingestor(req) {
     type = hashTypes[hash];
 
     try {
-      if (ensNormalize(message.space) !== message.space.toLowerCase()) throw new Error('');
+      if (
+        (message.space.startWith('s:') || !message.space.includes(':')) &&
+        ensNormalize(message.space) !== message.space.toLowerCase()
+      )
+        throw new Error('');
     } catch (e) {
       return Promise.reject('Invalid space id');
     }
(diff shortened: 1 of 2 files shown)


src/ingestor.ts
```diff
@@ -79,6 +80,16 @@ export default async function ingestor(req) {
     if (!Object.keys(hashTypes).includes(hash)) return Promise.reject('wrong types');
     type = hashTypes[hash];
 
+    try {
+      if (
+        (message.space.startsWith('s:') || !message.space.includes(':')) &&
+        ensNormalize(message.space) !== message.space.toLowerCase()
```
ChaituVR, line 86: I think v2 should send requests with just space id without the `s:` prefix

src/ingestor.ts
```diff
@@ -79,6 +80,16 @@ export default async function ingestor(req) {
     if (!Object.keys(hashTypes).includes(hash)) return Promise.reject('wrong types');
     type = hashTypes[hash];
 
+    try {
+      if (
+        (message.space.startsWith('s:') || !message.space.includes(':')) &&
```
ChaituVR, line 85: Not sure how we missed before but it should throw an error if `message.space` is undefined no?

src/ingestor.ts
ChaituVR, line 86: Maybe we can move to `getSpace`?

ChaituVR (changes requested):

src/ingestor.ts
wa0x6e: We can't, as it's required to differentiate different networks.

Sequencer also accepts space id like `sn:0x0000` and `sep:0x000`

src/ingestor.ts
wa0x6e: I believe the schema validation is already ensuring `message.space` is always set

src/ingestor.ts
wa0x6e: See https://github.com/snapshot-labs/snapshot-sequencer/pull/417#issuecomment-2255396997

src/ingestor.ts
ChaituVR: No 🙈 it is optional. but something else is preventing it not to throw an error 🤔

src/ingestor.ts
ChaituVR: lolol

src/ingestor.ts
ChaituVR: I will run few tests and see. but else it looks good, we can merge

ChaituVR: I tried to edit profile with this change (alias request) and it fail 
```ts
undefined
TypeError: Cannot read properties of undefined (reading 'startsWith')
```

ChaituVR: how it was working before 😅 with https://github.com/snapshot-labs/snapshot-sequencer/pull/417/files

wa0x6e: Maybe we could skip the check if `!message.space`

ChaituVR (approved):
tAck

PR #428, opened 2024-08-24
fix: Reduce proposal limit for flagged spaces

ChaituVR:
Context: 

https://discord.com/channels/955773041898573854/1103266859870081114/1276396777603334248

## Summary:
It should not be possible to create proposals on flagged spaces


wa0x6e (approved):
utAck

PR #437, opened 2024-09-21
fix: fix stamp cache refresh on profile name update not triggering

wa0x6e:
This PR fix the issue where the stamp cache refresh was not triggered on profile name update, due to monitoring on wrong field name


ChaituVR:
It looks like tests expect 1 call where it should be 2 😆

ChaituVR (approved):
tAck

wa0x6e:
Should work, still looking at the test, it should be called only once, because the other field did not change

PR #438, opened 2024-09-22
chore: fix tests

wa0x6e:
Fix tests from #437 

Tests have been fixed incorrectly in #437, this PR fix that. There should be only 1 call to `clearStampCache`

Also move the test to integration, since they depends on a database, and are not unit test


ChaituVR (approved):
utAck

PR #439, opened 2024-09-24
feat: ensure domain uniqueness

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/142
Together with https://github.com/snapshot-labs/snapshot-hub/pull/928

When saving space settings with domain, this PR will also save the domain in the `domain` column, and ensure that the domain is unique

See the hub PR for database update


src/helpers/actions.ts
```diff
@@ -7,19 +7,21 @@ export async function addOrUpdateSpace(space: string, settings: any) {
...
 
   await db.queryAsync(query, [
     {
       id: space,
       name: settings.name,
       created: ts,
       updated: ts,
-      settings: JSON.stringify(settings)
+      settings: JSON.stringify(settings),
+      domain: settings.domain || null
```
ChaituVR, line 19: i think better to be undefined 
```suggestion
      domain: settings.domain || undefined
```

src/writer/settings.ts
```diff
@@ -82,6 +82,17 @@ export async function verify(body): Promise<any> {
   const isAdmin = admins.includes(body.address.toLowerCase());
   const newAdmins = (msg.payload.admins || []).map(admin => admin.toLowerCase());
 
+  const anotherSpaceWithDomain = (
+    await db.queryAsync('SELECT 1 FROM spaces WHERE LOWER(domain) = LOWER(?) AND id != ? LIMIT 1', [
+      msg.payload.domain,
+      msg.space
+    ])
+  )[0];
```
ChaituVR, line 90: We should restrict it to be always lowercase here https://github.com/snapshot-labs/snapshot.js/blob/master/src/schemas/space.json#L104-L107

ChaituVR (changes requested):

src/helpers/actions.ts
wa0x6e: What will it do on the SQL side when passing undefined, and when there is previous value ? Will it just skip this column, or set undefined ? 

This column is only accepting either NULL or string, so what will undefined be ?

ChaituVR: ahh i was assuming it to be `domain` inside JSON 😆

src/writer/settings.ts
wa0x6e: See https://github.com/snapshot-labs/snapshot.js/pull/1060

ChaituVR:
Any idea about the test @wa0x6e

wa0x6e:
> Any idea about the test @wa0x6e

was unable to fix the tests, they're failing because of other tests, since the database is shared among all integrations, which are run in parallel

Running the test individually work (use `dotenv -e test/.env.test yarn jest test/integration/writer/settings.test.ts`)

This test is disabled for now, so we can move on

ChaituVR (approved):
tAck

PR #457, opened 2024-10-14
fix: convert to html statement to markdown

wa0x6e:


ChaituVR:
Are we sure that no one updated their statements? (re-importing statements from karma may override them)

wa0x6e:
> Are we sure that no one updated their statements? (re-importing statements from karma may override them)

We can check by comparing the created and updated date on statement with karmahq source

ChaituVR:
> > Are we sure that no one updated their statements? (re-importing statements from karma may override them)
> 
> We can check by comparing the created and updated date on statement with karmahq source

Looks like they are not updated

ChaituVR (approved):
tAck

PR #458, opened 2024-10-14
fix: remove statement source on statement update

wa0x6e:
This PR removes the `source` property on statement update


ChaituVR (approved):
utAck

PR #459, opened 2024-10-15
fix: Remove https from domain and default delegationNetwork to mainnet

ChaituVR:
Fixes https://github.com/snapshot-labs/workflow/issues/253

## How to test

- In `src/helpers/poke.ts` comment the code that checks validation: (until we merge https://github.com/snapshot-labs/snapshot.js/pull/1074)
```ts
// if (snapshot.utils.validateSchema(snapshot.schemas.space, space) !== true) {
    //   return Promise.reject('invalid space format');
    // }
```
- Open http://localhost:3001/spaces/cow.eth/poke
- Should save space settings without any error (domain should be saved as `vote.cow.fi`)
- In `src/helpers/actions.ts` add `delete settings.domain` on line `10`
- Open http://localhost:3001/spaces/cowtesting.eth/poke 
- Should save space settings without any error

PR #462, opened 2024-10-17
refactor: delegationNetwork default value

ChaituVR:
as suggested https://github.com/snapshot-labs/snapshot-sequencer/pull/459#discussion_r1802752060


wa0x6e (approved):
tAck

PR #469, opened 2024-11-04
fix: add new columns in schema.sql

### LINKED ISSUE #468: fix: update schema.sql to reflect new columns
Schema.sql is unsynced with database real schema, with some column such as this one not existing on the sql file

https://github.com/snapshot-labs/snapshot-sequencer/blob/4908c0bfff6c338095a50f26d194cec9cdb452f9/src/writer/proposal.ts#L249

Which will throw

```
ERROR_BAD_FIELD_ERROR: Unknown column 'vp_value_by_strategy' in 'field list'
```

when running the test

wa0x6e:
Fix #468 

Updated schema.sql proposals to match with production database state

Added missing columns:

- `scores_total_value`
- `vp_value_by_strategy`

Added missing `DEFAULT` to columns with default

Added missing INDEX

Added notes about some discrepancies with production database schema


wa0x6e:
Any updates on this one ? All tests are failing

ChaituVR:
As far as I remember @bonustrack asked us to wait, but I'm not sure exactly. Do you think we can proceed? @bonustrack

bonustrack:
I've changed the new types on the `proposals` table to be:
image
If you can update it, then all good for me. Later I'll also add some types on the `votes` table

wa0x6e:
Updated

test/schema.sql
```diff
@@ -29,15 +29,15 @@ CREATE TABLE spaces (
 );
 
 CREATE TABLE proposals (
-  id VARCHAR(66) NOT NULL,
-  ipfs VARCHAR(64) NOT NULL,
+  id VARCHAR(66) NOT NULL DEFAULT '',
```
ChaituVR, line 32: Don't think we should have default values, specially with `id` we always want it to be random
any reason to add default values?

also not sure we want to update our prod DB with these 🙈  better to change schema of new values only? 😅

test/schema.sql
```diff
@@ -49,17 +49,20 @@ CREATE TABLE proposals (
...
   scores_by_strategy JSON NOT NULL,
-  scores_state VARCHAR(24) NOT NULL,
+  scores_state VARCHAR(24) NOT NULL DEFAULT '',
   scores_total DECIMAL(64,30) NOT NULL,
   scores_updated INT(11) NOT NULL,
+  scores_total_value DECIMAL(64,30) NOT NULL DEFAULT '0.000000000000000000000000000000',
+  vp_value_by_strategy json NOT NULL,
   votes INT(12) NOT NULL,
   flagged INT NOT NULL DEFAULT 0,
+  cb INT NOT NULL DEFAULT 0,
```
ChaituVR, line 65: No cb on We should add same on https://github.com/snapshot-labs/snapshot-hub/blob/master/src/helpers/schema.sql 🤔 new field?

test/schema.sql
ChaituVR: hmmm correct 😢 werid, i don't understand why we need a default, maybe production has a outdated schema 🥲

ChaituVR: which way we go?

test/schema.sql
ChaituVR, line 1: We should keep it the same as hub's schema.sql to avoid confusion, we should change it there along with this

test/schema.sql
ChaituVR: I don't think `id` should have a default value, because this should be unique, and we can't update prod DB now to remove default (it may take lot of time). maybe we can just remove this default?

test/schema.sql
```diff
@@ -29,15 +29,15 @@ CREATE TABLE spaces (
 );
 
 CREATE TABLE proposals (
-  id VARCHAR(66) NOT NULL,
-  ipfs VARCHAR(64) NOT NULL,
+  id VARCHAR(66) NOT NULL DEFAULT '',
+  ipfs VARCHAR(64) NOT NULL DEFAULT '',
```
ChaituVR, line 33: default here also doesn't make sense we always need an IPFS

ChaituVR (changes requested):
We

test/schema.sql
wa0x6e: Will open a PR on the hub with the same changes once this PR is approved

test/schema.sql
```diff
@@ -49,17 +53,20 @@ CREATE TABLE proposals (
...
   privacy VARCHAR(24) NOT NULL,
   snapshot INT(24) NOT NULL,
   app VARCHAR(24) NOT NULL,
   scores JSON NOT NULL,
   scores_by_strategy JSON NOT NULL,
-  scores_state VARCHAR(24) NOT NULL,
+  scores_state VARCHAR(24) NOT NULL DEFAULT '',
   scores_total DECIMAL(64,30) NOT NULL,
   scores_updated INT(11) NOT NULL,
+  scores_total_value DECIMAL(64,30) NOT NULL DEFAULT '0.000000000000000000000000000000',
```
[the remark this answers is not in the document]
ChaituVR: I feel like this default size can increase our DB size no?

PR #470, opened 2024-11-12
feat: add email subscriptions

wa0x6e:


wa0x6e:
Tests are failing for issue unrelated to this PR, should he fixed via https://github.com/snapshot-labs/snapshot-sequencer/pull/469

test/schema_envelop.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE subscribers (
+  email VARCHAR(256) NOT NULL,
+  address VARCHAR(256) NOT NULL,
+  subscriptions JSON DEFAULT NULL,
+  created BIGINT NOT NULL,
+  verified BIGINT NOT NULL DEFAULT 0,
+  PRIMARY KEY (email, address),
```
ChaituVR, line 7: The primary key can be just `address` no?  because a user cannot use multiple emails. this will help us to use queries like `INSERT ... ON DUPLICATE KEY UPDATE` to insert or update email subscription in   `src/writer/email-subscription.ts`

ChaituVR (changes requested):

test/schema_envelop.sql
wa0x6e: An address can be associated to multiple email addresses, but only one can be verified.

I think this was done so an unverified address can subscribe again with another email address

PR #475, opened 2024-12-05
fix: Allow `shutter` in proposal privacy if space is using `shutter` privacy

ChaituVR:
Discussion https://github.com/snapshot-labs/sx-monorepo/pull/954#discussion_r1867920957

### Summary

- Modifies privacy check to Allow `shutter` in proposal privacy if space is using `shutter` privacy

### How to test

- Use UI of https://github.com/snapshot-labs/sx-monorepo/pull/954
- Try changing privacy settings to `shutter`
- Try creating a proposal 
- Proposal should go in without any error (also with old UI)

### DIFF
--- a/src/writer/update-proposal.ts
+++ b/src/writer/update-proposal.ts
@@ -49,11 +49,10 @@ export async function verify(body): Promise<any> {
 
   if (proposal.author !== body.address) return Promise.reject('Not the author');
 
-  if (
-    space.voting?.privacy !== 'any' &&
-    msg.payload.privacy &&
-    !(space.voting.privacy === 'shutter' && msg.payload.privacy === 'shutter')
-  ) {
+  const spacePrivacy = space.voting?.privacy ?? 'any';
+  const proposalPrivacy = msg.payload.privacy;
+
+  if (proposalPrivacy !== undefined && spacePrivacy !== 'any' && spacePrivacy !== proposalPrivacy) {
     return Promise.reject('not allowed to set privacy');
   }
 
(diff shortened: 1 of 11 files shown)


wa0x6e (changes requested):
Added some tests, and seems like 2 scenarios are failing

ChaituVR:
Fixed tests

test/unit/writer/proposal.test.ts
```diff
@@ -425,11 +425,11 @@ describe('writer/proposal', () => {
           expect(mockGetSpace).toHaveBeenCalledTimes(1);
         });
 
-        it('rejects a proposal with no privacy', async () => {
+        it('accepts a proposal with empty string in privacy', async () => {
```
wa0x6e, line 428: This test is nested inside `'when the space is using SHUTTER privacy'`, and should accept only `shutter` as value, and not `""`

test/schema.sql
```diff
@@ -58,6 +58,7 @@ CREATE TABLE proposals (
   scores_state VARCHAR(24) NOT NULL,
   scores_total DECIMAL(64,30) NOT NULL,
   scores_updated INT(11) NOT NULL,
+  vp_value_by_strategy JSON DEFAULT NULL,
```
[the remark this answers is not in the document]
ChaituVR:
```suggestion
```

ChaituVR:
Tests failing are not related to this PR, we can proceed without it

test/unit/writer/proposal.test.ts
wa0x6e: The test above is for another scenario.

This current test is for when the space has set `SHUTTER` as privacy, and we're sending an empty string as proposal privacy.

In this case, the writer should reject, since empty string (== no privacy), and we can not disable privacy is space has enforced it.

src/writer/update-proposal.ts
```diff
@@ -49,7 +49,11 @@ export async function verify(body): Promise<any> {
 
   if (proposal.author !== body.address) return Promise.reject('Not the author');
 
-  if (space.voting?.privacy !== 'any' && msg.payload.privacy) {
+  if (
+    space.voting?.privacy !== 'any' &&
+    msg.payload.privacy &&
```
wa0x6e, line 54: This check will ignore empty string, and is the reason why one of the original test is failing

wa0x6e (approved):
tAck

PR #486, opened 2025-01-06
fix: ignore deleted spaces when checking for followed spaces limit

wa0x6e:
Following on https://github.com/snapshot-labs/snapshot-sequencer/pull/401#issuecomment-2210284601

This PR fix an edge case when checking the limit of followed spaces, when a followed space has been deleted, but will still be counted toward the limit.

This PR will ignore deleted spaces from the limit count

### DIFF
--- a/src/writer/follow.ts
+++ b/src/writer/follow.ts
@@ -3,7 +3,12 @@ import db from '../helpers/mysql';
 import { DEFAULT_NETWORK_ID, NETWORK_IDS } from '../helpers/utils';
 
 export const getFollowsCount = async (follower: string): Promise<number> => {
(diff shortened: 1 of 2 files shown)


wa0x6e:
Some tests are failing due to issues unrelated to this PR, see https://github.com/snapshot-labs/snapshot-sequencer/pull/469

ChaituVR:
Is there an example to test?

ChaituVR (approved):
utAck

wa0x6e:
> Is there an example to test?

Not easily, you have to connect to a local db, and manually flag spaces as deleted, or just trust the integrations tests

PR #487, opened 2025-01-06
feat: sync spaces custom domain with walletconnect allowed origins

wa0x6e:
Closes #https://github.com/snapshot-labs/workflow/issues/362 

This PR will sync the space's custom domain with WalletConnect allowed origins list, using walletconnect's new API

## Changes

- Adding a custom domain will also add on walletconnect
- Removing a custom domain will remove it from walletconnect
- Editing a custom domain will add the new one, and remove the old one from walletconnect

## How to test

Set the new env vars in your .env

```
WALLETCONNECT_PROJECT_ID= # optional
REOWN_SECRET= # optional
```

You can create your own (via https://cloud.reown.com/app), or ask for the prod keys

- Connect an UI (e.g sx or v1) to sequencer
- On your terminal, send `curl -X GET "https://cloud.reown.com/api/get-allowed-domains?projectId=[WALLETCONNECT_PROJECT_ID]" -H "Authorization: [REOWN_SECRET]"` (set the correct variables)
- It should return a list of whitelisted domains
- On the UI, adds a new custom domain to your space
- Send the curl command
- The returned list should include your newly added domain
- On the UI, update your space's custom domain
- Send the curl command again
- Your previous domain should be removed, and replaced by your updated one
- On the UI, remove the custom domain from your space
- Send the curl command again
- Your domain should be removed from the list


wa0x6e:
We currently have 1033 custom domain, and most of them are not working.

Should we import all of them, or import only those working ?

src/writer/settings.ts
```diff
@@ -133,4 +140,9 @@ export async function action(body): Promise<void> {
   }
 
   if (existingSettings.avatar !== msg.payload.avatar) await clearStampCache('space', space);
+
+  if (existingSpace?.domain != msg.payload.domain) {
+    await addToWalletConnectWhitelist(msg.payload.domain);
```
ChaituVR, line 145: I am still wondering if we want to add all domains (any space to add it) or just turbo spaces

wa0x6e: We need to decide whether we allow `domain` update for all spaces, or just turbo

bonustrack: If that's simpler we can allow only for turbo and the ones who already had a domain set to update domain.

wa0x6e: We need to set boundaries on the definition of "already had a domain".

For now, if a non-turbo space has already set a domain, he can change it.

But what happen when a non-turbo space currently has a domain, but remove it (by error, etc ...) ? The way it work now, the space lose the ability to put it back. To enable this, we need a way to mark these spaces as custom domain enabled

ChaituVR: All i worry about this PR is that someone may add many unwanted domains to walletconnect as any space can add a domain

ChaituVR: Otherwise looks good

wa0x6e: > All i worry about this PR is that someone may add many unwanted domains to walletconnect as any space can add a domain

Working on a PR to prevent that. domain edition will be limited to turbo, and spaces with existing domains only

wa0x6e: See https://github.com/snapshot-labs/snapshot-sequencer/pull/488

This PR can be merged after the other one

.env.example
```diff
@@ -16,3 +16,5 @@ BROVIDER_URL=https://rpc.snapshot.org # optional
 STARKNET_RPC_URL= # optional
 # Secret for laser
 AUTH_SECRET=1dfd84a695705665668c260222ded178d1f1d62d251d7bee8148428dac6d0487 # optional
+WALLET_CONNECT_PROJECT_ID= # optional
```
bonustrack, line 19:
```suggestion
WALLETCONNECT_PROJECT_ID= # optional
```

wa0x6e:
What's stalling with this PR ?

bonustrack (approved):
tACK

bonustrack:
I've added the env on Sequencer mainnet and testnet

wa0x6e:
> I've added the env on Sequencer mainnet and testnet

Not sure we need that for testnet, as whitelabel does not work there

ChaituVR:
> > I've added the env on Sequencer mainnet and testnet
> 
> Not sure we need that for testnet, as whitelabel does not work there

I've removed these two on testnet

ChaituVR (approved):
utAck

src/helpers/utils.ts
```diff
@@ -212,3 +212,38 @@ export function captureError(e: any, context?: any, ignoredErrorCodes?: number[]
...
+    await fetch(`https://cloud.reown.com/api/set-allowed-domains`, {
+      method,
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        Authorization: process.env.REOWN_SECRET
+      },
+      body: JSON.stringify({
+        projectId: process.env.WALLETCONNECT_PROJECT_ID,
+        origins: [`https://${domain}`]
```
bonustrack, line 240: We don't need `https://` here
```suggestion
        origins: [domain]
```

bonustrack:
@wa0x6e Can you fix the test?

PR #490, opened 2025-01-10
feat: add script to import skins

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/275

This PR adds a script to import skins from the snapshot-spaces repository into the new `skins` table.

Only skins used by the v2 (skins used by turbo spaces, and spaces with existing custom domains) will be imported

Create the table with

```
CREATE TABLE skins (
  id VARCHAR(100) NOT NULL,
  bg_color VARCHAR(6) DEFAULT NULL,
  link_color VARCHAR(6) DEFAULT NULL,
  text_color VARCHAR(6) DEFAULT NULL,
  content_color VARCHAR(6) DEFAULT NULL,
  border_color VARCHAR(6) DEFAULT NULL,
  heading_color VARCHAR(6) DEFAULT NULL,
  primary_color VARCHAR(6) DEFAULT NULL,
  header_color VARCHAR(6) DEFAULT NULL,
  PRIMARY KEY (id)
);
```

### DIFF
--- a/scripts/import-skins.ts
+++ b/scripts/import-skins.ts
@@ -233,6 +233,14 @@ async function loadAndConvertSkin(skin: string) {
       colors[`${key}_color`] = color;
     });
 
+    if (colors['text_color']) {
+      colors['content_color'] = colors['text_color'];
+      const textRgba = hexToRgba(colors['text_color']);
+      textRgba[3] = 0.85;
+
+      colors['text_color'] = opacifyColor(textRgba, hexToRgba(colors['bg_color'] || 'ffffff'));
+    }
+
     skins[skin] = colors;
   } catch (e) {
     console.log(e);
@@ -272,8 +280,8 @@ async function main() {
 
       return db.queryAsync(
         `
-          INSERT INTO skins (id, bg_color, link_color, text_color, border_color, heading_color, primary_color, header_color)
-          VALUES(?, ?, ?, ?, ?, ?, ?, ?)
+          INSERT INTO skins (id, bg_color, link_color, text_color, content_color, border_color, heading_color, primary_color, header_color)
+          VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)
           ON DUPLICATE KEY UPDATE id=id
         `,
         [
@@ -281,6 +289,7 @@ async function main() {
           skin.bg_color,
           skin.link_color,
           skin.text_color,
+          skin.content_color,
           skin.border_color,
           skin.heading_color,
           skin.primary_color,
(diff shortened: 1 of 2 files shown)


ChaituVR:
@wa0x6e This one is ready to review? 🤔

scripts/import-skins.ts
```diff
@@ -0,0 +1,308 @@
...
+      settings['content_color'] = settings['text_color'];
+      const textRgba = hexToRgba(settings['text_color']);
+      textRgba[3] = 0.85;
+
+      settings['text_color'] = opacifyColor(textRgba, hexToRgba(settings['bg_color'] || 'ffffff'));
+    }
+
+    if (settings['bg_color']) {
+      const [r, g, b] = hexToRgba(settings['bg_color']);
+      settings['theme'] = getBrightness(r, g, b) > 0.5 ? 'light' : 'dark';
```
ChaituVR, line 232: Is it required to store `theme` if it is derived from bg_color? 🤔

test/schema.sql
```diff
@@ -193,6 +193,20 @@ CREATE TABLE leaderboard (
   INDEX last_vote (last_vote)
 );
 
+CREATE TABLE skins (
+  id VARCHAR(100) NOT NULL,
```
ChaituVR, line 197: If this is space id, we can remove `skin` in `spaces` right?

scripts/import-skins.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  const spacesWithCustomDomain = await db.queryAsync(`
+    SELECT
+      id,
+      JSON_UNQUOTE(settings->'$.skin') as skin
+    FROM spaces
+    WHERE
+      settings->'$.skin' IS NOT NULL
+      AND domain IS NOT NULL
+  `);
```
ChaituVR, line 252: Doesn't need turbo at this stage?

test/schema.sql
wa0x6e: It's being left there for backward compatibility

scripts/import-skins.ts
wa0x6e: `AND domain IS NOT NULL` is the filter. Skin is enabled on all whitelabel

scripts/import-skins.ts
wa0x6e: It's derived from background color when importing, since it didn't exist on v1.

Later, we may allow users to edit this setting

ChaituVR (approved):
tAck

PR #495, opened 2025-01-30
feat: migrate turbo limits checks from SDK

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/315
Depends on https://github.com/snapshot-labs/snapshot-sequencer/pull/494

After removing the turbo limits check in https://github.com/snapshot-labs/snapshot.js/pull/1115 , this PR will add back those checks on sequencer side.

It will check for 

- proposal body length
- number of proposal choices
- number of space strategies


src/writer/proposal.ts
```diff
@@ -217,6 +208,16 @@ export async function verify(body): Promise<any> {
...
   }
+
+  const bodyLengthLimit = await getLimit(`space.${await getSpaceType(space)}.body_limit`);
+  if (msg.payload.body.length > bodyLengthLimit) {
+    return Promise.reject(`proposal body length can not exceed ${bodyLengthLimit} characters`);
+  }
+
+  const choicesLimit = await getLimit(`space.${await getSpaceType(space)}.choices_limit`);
+  if (msg.payload.choices.length > choicesLimit) {
+    return Promise.reject(`number of choices can not exceed ${choicesLimit}`);
```
ChaituVR, line 219: I think we should call `getSpaceType` on top level instead of four times, we can then pass `spaceType` to `getSpaceProposalsLimits` instead of `space`

ChaituVR: maybe we should also have `isEcosystemSpace` on top and pass type to `getSpaceProposalsLimits`

ChaituVR (changes requested):

ChaituVR (approved):
utAck

will test in the other PR

ChaituVR:
@wa0x6e Can we merge this? this is blocking all other PRs

PR #497, opened 2025-02-03
feat: update space skins settings

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/418
Depends on https://github.com/snapshot-labs/snapshot.js/pull/1119

This PR will add support for updating the skins table when colors are passed in the space settings payload, only for whitelabel spaces

- When `skinParams` is missing, will not do anything on the skins
- When passed, will creates or update the existing skin 

When some skin properties are skipped, it will fallback to null, and will delete existing value


ChaituVR:
we handle these on https://github.com/snapshot-labs/snapshot-sequencer/pull/498 right? we can close this maybe?

wa0x6e:
> we handle these on #498 right? we can close this maybe?

The other one is for adding a the custom logo only. The list of changed files will update once this one is merged into master

src/helpers/actions.ts
```diff
@@ -2,36 +2,81 @@ import snapshot from '@snapshot-labs/snapshot.js';
...
   await db.queryAsync(query, [
     {
-      id: space,
+      id,
       name: settings.name,
       created: ts,
       updated: ts,
-      settings: JSON.stringify(settings),
+      settings: JSON.stringify(normalizedSettings),
       domain: settings.domain || null
```
ChaituVR, line 39: I think we should use `normalizedSettings.domain` here also right? 

is it because  `settings.domain` for new space will always be `undefined` considering it won't be turbo while creation 😄 then its fine

src/helpers/actions.ts
ChaituVR, line 1: Don't see an easy way to test skinSettings without UI 😅  maybe can I do a utAck

ChaituVR (approved):
tAck

except not sure about `domain` field (above comment)

src/helpers/actions.ts
wa0x6e: I test sequencer without UI, just sending SDK requests with various payload, and testing whether data get saved in the skins table

PR #498, opened 2025-02-04
feat: add support for new skin logo in space settings

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/416
Depends on https://github.com/snapshot-labs/snapshot.js/pull/1120

This PR adds support for saving and updating the logo passed in `skinParams`

Create the new required columns with 
```sql
ALTER TABLE skins
ADD COLUMN logo VARCHAR(256) DEFAULT NULL AFTER theme;
```


src/helpers/actions.ts
```diff
@@ -2,36 +2,95 @@ import snapshot from '@snapshot-labs/snapshot.js';
...
+    'border_color',
+    'heading_color',
+    'primary_color',
+    'header_color'
+  ];
+
+  const _params = snapshot.utils.clone(skinParams);
+  COLORS.forEach(color => {
+    if (_params[color]) {
+      _params[color] = _params[color].replace('#', '');
```
ChaituVR, line 69: feel it would be easy to just store # in DB instead of replacing it and adding it again

src/helpers/actions.ts
```diff
@@ -2,36 +2,95 @@ import snapshot from '@snapshot-labs/snapshot.js';
...
     },
     ts,
-    JSON.stringify(settings),
+    JSON.stringify(normalizedSettings),
     settings.name,
-    settings.domain || null
+    normalizedSettings.domain || null
   ]);
+
+  await addOrUpdateSkin(id, settings.skinParams);
```
ChaituVR, line 49: Not sure if it is good to call `addOrUpdateSkin` inside `addOrUpdateSpace`, maybe we should just call it after `addOrUpdateSpace` everywhere

src/helpers/actions.ts
ChaituVR, line 69: `_params[color]` can be undefined right? these are not mandatory fields in snapshot.js

src/helpers/actions.ts
```diff
@@ -2,36 +2,95 @@ import snapshot from '@snapshot-labs/snapshot.js';
...
-      settings: JSON.stringify(settings),
+      settings: JSON.stringify(normalizedSettings),
       domain: settings.domain || null
     },
     ts,
-    JSON.stringify(settings),
+    JSON.stringify(normalizedSettings),
     settings.name,
-    settings.domain || null
+    normalizedSettings.domain || null
```
ChaituVR, line 46: We store `domain` in both `settings` and `domain` field? 🤔

src/helpers/actions.ts
```diff
@@ -2,36 +2,95 @@ import snapshot from '@snapshot-labs/snapshot.js';
...
+    [
+      {
+        id,
+        ..._params
+      },
+      ...COLORS.map(color => _params[color]),
+      _params.theme || existingSkin?.theme || DEFAULT_SKIN,
+      _params.logo || existingSkin?.logo || null
+    ]
+  );
```
ChaituVR, line 93: We can use existing values without need to call DB
```suggestion
  await db.queryAsync(
  `INSERT INTO skins
    SET ?
    ON DUPLICATE KEY UPDATE
      ${COLORS.map(color => `${color} = COALESCE(VALUES(${color}), ${color})`).join(', ')},
      theme = COALESCE(VALUES(theme), theme),
      logo = COALESCE(VALUES(logo), logo)
  `,
  [
    {
      id,
      ..._params
    },
    ...COLORS.map(color => _params[color]),
    _params.theme,
    _params.logo
  ]
);
```

ChaituVR (changes requested):

ChaituVR:
I ran this on prod DB and testnet DB

```SQL
ALTER TABLE skins
ADD COLUMN logo VARCHAR(256) DEFAULT NULL AFTER theme;
```

src/helpers/actions.ts
wa0x6e: Yes

src/helpers/actions.ts
wa0x6e: Maybe we should stop using fallback for logo, this prevents sending null, which mean remove the existing logo

If you don't send a value, it will mean you want to delete it (like in colors)

src/helpers/actions.ts
wa0x6e:
```sql 
UPDATE skins 
SET bg_color = CONCAT('#', bg_color)
WHERE bg_color IS NOT NULL;

UPDATE skins 
SET link_color = CONCAT('#', link_color)
WHERE link_color IS NOT NULL;

UPDATE skins 
SET text_color = CONCAT('#', text_color)
WHERE text_color IS NOT NULL;

UPDATE skins 
SET border_color = CONCAT('#', border_color)
WHERE border_color IS NOT NULL;

UPDATE skins 
SET heading_color = CONCAT('#', heading_color)
WHERE heading_color IS NOT NULL;

UPDATE skins 
SET primary_color = CONCAT('#', primary_color)
WHERE primary_color IS NOT NULL;

UPDATE skins 
SET header_color = CONCAT('#', header_color)
WHERE header_color IS NOT NULL;
```

Run this at the time of merging this PR

src/helpers/actions.ts
wa0x6e: `addOrUpdateSkin` is dependent of the `addOrUpdateSpace`, and will never be called without it. Why would making an extra call everywhere better ?

ChaituVR (approved):
utAck

ChaituVR:
> > utAck
> 
> Is the new column added to the table ?

Yes ser https://github.com/snapshot-labs/snapshot-sequencer/pull/498#issuecomment-2635815169

PR #502, opened 2025-02-14
feat: add Mantle support for follows

Sekhmet:


ChaituVR (approved):
utAck

Sekhmet:
@ChaituVR not sure why but tests fail with schema validation, looks like it doesn't use latest snapshot.js here 🤔

ChaituVR:
Hmm it happens randomly. not sure why. triggered a rerun, should be fine now

PR #507, opened 2025-02-26
fix:  `delete-space` action should use default network instead of space network

ChaituVR:
https://github.com/snapshot-labs/sx-monorepo/pull/1218#discussion_r1971371127

This is to allow only mainnet Safes to delete spaces, similar to `settings` action


wa0x6e (approved):
utAck

PR #511, opened 2025-03-07
chore: log unhandled errors

ChaituVR:
- Will be helpful to debug

Sometimes logs are not printed if error is not a object,

Try passing `plugins: ""` while proposal creation,
Before:
```bash
warn: [ingestor] msg validation failed (typed data): {}
```

After
```bash
warn: [ingestor] msg validation failed (typed data): Unexpected end of JSON input
```

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -25,7 +25,7 @@ router.post(
       const result = await typedData(req);
       res.json(result);
     } catch (e: any) {
-      const errorMessage = typeof e === 'object' ? e?.message || JSON.stringify(e) : String(e);
+      const errorMessage = typeof e === 'object' ? e.message || JSON.stringify(e) : String(e);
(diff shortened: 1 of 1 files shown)


src/api.ts
```diff
@@ -24,8 +24,9 @@ router.post(
     try {
       const result = await typedData(req);
       res.json(result);
-    } catch (e) {
-      log.warn(`[ingestor] msg validation failed (typed data): ${JSON.stringify(e)}`);
+    } catch (e: any) {
+      const errorMessage = typeof e === 'object' ? e?.message || JSON.stringify(e) : String(e);
```
wa0x6e, line 28: why `e?.message` when `e` is already checked for existence with `typeof` ?

ChaituVR:
```suggestion
      const errorMessage = typeof e === 'object' ? e.message || JSON.stringify(e) : String(e);
```

PR #512, opened 2025-03-07
fix: accept empty string/object in plugins

ChaituVR:
Whilte creating proposals, Our docs suggest that we need to pass `plugins`  as object in string
```js
plugins:  JSON.stringify({})
```

but many times users send a empty string or empty object for `plugins`
```js
plugins:  ''
```

```js
plugins:  {}
```

This fix will accept both and fallback to `{}`


wa0x6e (approved):
utAck

PR #513, opened 2025-03-09
feat: validate space strategies

wa0x6e:


src/writer/settings.ts
```diff
@@ -80,6 +81,26 @@ export async function verify(body): Promise<any> {
     return Promise.reject(`max number of strategies is ${strategiesLimit}`);
   }
 
+  try {
+    const strategiesList = await (await fetchWithKeepAlive(`${scoreAPIUrl}/api/strategies`)).json();
+
+    msg.payload.strategies
+      .map(strategy => strategy.name)
```
ChaituVR, line 88: there is also one more level of strategies,for example using `multichain` inside `delegation` 🙈

src/writer/settings.ts
```diff
@@ -80,6 +81,26 @@ export async function verify(body): Promise<any> {
     return Promise.reject(`max number of strategies is ${strategiesLimit}`);
   }
 
+  try {
+    const strategiesList = await (await fetchWithKeepAlive(`${scoreAPIUrl}/api/strategies`)).json();
```
ChaituVR, line 85: Maybe we should call it every 5 mins instead of calling it everytime, but can make things look complicated 🙈

src/writer/settings.ts
```diff
@@ -80,6 +81,26 @@ export async function verify(body): Promise<any> {
     return Promise.reject(`max number of strategies is ${strategiesLimit}`);
   }
 
+  try {
```
ChaituVR, line 84: Maybe we should move this check and strategies limit check to a new function similar to `validateSpaceSettings`

ChaituVR (changes requested):

ChaituVR (dismissed):
utAck, other than above this looks good

wa0x6e:
The PR had been refactored since last review, to fetch and cache the strategies list, like on the hub, instead of getting it on each `verify()` call.

Also some refactoring to extract `validateSpaceSetting` into a helper, for easier testing.

See the updated PR description

ChaituVR (changes requested):

src/writer/settings.ts
```diff
@@ -82,12 +23,15 @@ export async function verify(body): Promise<any> {
...
-    });
+    await validateSpaceSettings(
+      {
+        ...msg.payload,
+        id: msg.space,
+        deleted: space?.deleted,
+        turbo: space?.turbo
+      },
+      process.env.NETWORK
+    );
```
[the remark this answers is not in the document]
wa0x6e: This is to have a pure function in strategies, where all variables are defined in a single place (the arguments). I am using a the Dependency injection pattern here, which allow for easier testing.

Trying to have less `process.env` scattered in the code as possible, because they're not always consistent, we have to remember to set the fallback (`|| 'testnet'`) on each file, which is hard to maintain.

What I envision for later, is to have a single constants.ts file, where we define the app global constants, and just import them to use, and not care about setting fallback in each file anymore, like:

```ts
// constants.ts
export SCORE_API_URL = process.env.SCORE_API_URL
```

```ts
// strategies.ts
import { SCORE_API_URL } from 'constants.ts'

// Then just use, no more need for the || https://score.snapshot.org each time
fetch(SCORE_API_URL) 
```

Some constants like the brovider url is repeated in 5 different files.

What do you think ?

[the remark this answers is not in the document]
ChaituVR: nice idea yea 👍 

but also we should just default second param of validateSpaceSettings to `process.env.NETWORK` instead of passing it everywhere, because it is just for testing

src/index.ts
```diff
@@ -8,25 +8,62 @@ import initMetrics from './helpers/metrics';
...
-app.disable('x-powered-by');
-app.use(express.json({ limit: '20mb' }));
-app.use(express.urlencoded({ limit: '20mb', extended: false }));
-app.use(cors({ maxAge: 86400 }));
-app.use(rateLimit);
-app.set('trust proxy', 1);
-app.use('/', api);
-app.use('/shutter', shutter);
+  await initializeStrategies();
+  refreshStrategies();
```
ChaituVR, line 25: Not a biggie but this send two requests back to back

ChaituVR: maybe we can just move await to top
```ts
export async function run() {
  while (!shouldStop) {
    await snapshot.utils.sleep(RUN_INTERVAL);
    try {
      log.info('[strategies] Start strategies refresh');
      await loadStrategies();
      consecutiveFailsCount = 0;
      log.info('[strategies] End strategies refresh');
    } catch (e: any) {
      consecutiveFailsCount++;

      if (consecutiveFailsCount >= MAX_CONSECUTIVE_FAILS) {
        capture(e);
      }
      log.error(`[strategies] failed to load ${JSON.stringify(e)}`);
    }
  }
}
```

PR #516, opened 2025-03-11
chore: increase follow limit for Turbo users

bonustrack:
Increase follow limit for Turbo users to 50.


src/helpers/turboUsers.json
```diff
@@ -0,0 +1,4 @@
+[
+  "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
+  "0x3FB19771947072629C8EEE7995a2eF23B72d4C8A"
+]
```
ChaituVR, line 4: We should move to DB similar to ecosystem spaces

PR #520, opened 2025-03-24
feat: use unified API to check if SX space exists

Sekhmet:
With this change we will use unified API to get space info.

I tested that `sxSpaceExists` tests pass.


bonustrack (approved):
utACK

wa0x6e (approved):
utAck

PR #521, opened 2025-03-24
feat: add mnt, ape, curtis to networks whitelist

Sekhmet:
Updated mantle to mnt as it's new NetworkID for it.


ChaituVR (approved):
utAck

PR #522, opened 2025-03-25
fix: prevent non-premium networks

ChaituVR:
Closes https://github.com/snapshot-labs/workflow/issues/497

## Summary
Stops proposal creation for spaces using non-premium network (main network, network inside strategy or inside multichain strategy)

### How to test
- Try changing the settings of space to use a non-premium network
- Create a proposal on v1
- It should throw any error
image


wa0x6e (approved):
tAck

PR #523, opened 2025-03-25
feat: add turbo tracking

pscott:
See: https://github.com/snapshot-labs/workflow/issues/488#issue-2902688159

- Adds a job that iterates every 10 seconds and pulls **all** the spaces that have `turbo` from schnaps.
- Based on the aforementioned list, updates the `turbo` and `turbo_expiration_timestamp` entries on the hub's DB 
- Adds a `SCHNAPS_API_URL` to the `.env.example`

This does not scale particularly well but was suggested by @bonustrack for a quick way of getting to production.
Later on, we will probably:
- Store somewhere the latest "synced" timestamp
- Only query from SCHNAPS the payments that were done since the last "synced" timestamp, and update accordinly

### DIFF
--- a/src/helpers/turbo.ts
+++ b/src/helpers/turbo.ts
@@ -44,14 +44,17 @@ async function updateTurboStatuses(spaces: { id: string; turbo_expiration: numbe
       turbo_expiration = CASE id
         ${spaces.map(() => `WHEN ? THEN ?`).join(' ')}
       END
-    WHERE id IN (${spaces.map(() => '?').join(', ')});
+    WHERE id IN (?);
   `;
 
   // Flatten `spaces` array: [id1, expiration1, id2, expiration2, ..., now, id1, id2, ...]
-  const params = spaces.flatMap(({ id, turbo_expiration }) => [id, turbo_expiration]);
+  const params: (string | string[] | number)[] = spaces.flatMap(({ id, turbo_expiration }) => [
+    id,
+    turbo_expiration
+  ]);
 
   // Append `id`s again for the `WHERE IN (...)` clause
-  params.push(...spaces.map(space => space.id));
+  params.push(spaces.map(space => space.id));
 
   await db.query(query, params);
 }


wa0x6e:
Do we have a dedicated instance for testnet ?

pscott:
> Do we have a dedicated schnaps instance for testnet ?

No. You can run it locally using `INDEX_TESTNET=true` to have it index `sepolia`.

Otherwise maybe @bonustrack can deploy something on DO?

src/helpers/turbo.ts
```diff
@@ -0,0 +1,90 @@
+import snapshot from '@snapshot-labs/snapshot.js';
+import fetch from 'node-fetch';
+import db from './mysql';
+
+const SCHNAPS_API_URL = process.env.SCHNAPS_API_URL || 'https://schnaps.snapshot.box';
+const NO_SCHNAPS = process.env.NO_SCHNAPS === 'true';
```
ChaituVR, line 6: Could we default this to true if `process.env.SCHNAPS_API_URL` is empty? This will make both these env optional. good for new devs, else not important

.env.example
```diff
@@ -9,6 +9,7 @@ SHUTTER_URL=https://...
 SIDEKICK_URL=https://sh5.co
 PINEAPPLE_URL=https://.../upload # optional
 SCORE_API_URL=https://score.snapshot.org
+SCHNAPS_API_URL='https://schnaps.snapshot.box'
```
ChaituVR, line 12:
```suggestion
SCHNAPS_API_URL=https://schnaps.snapshot.box
```

ChaituVR (commented):
Looks good, need to test it

but looks like https://schnaps.snapshot.box may add turbo to `test.wa0x6e.eth` 😄 we will use different URL on prod?

src/helpers/turbo.ts
pscott: good idea.
removed the default `SCHNAPS_API_URL`, removed `NO_SCHNAPS`, simplified the code: c1136b9e618f6d8b5f1e462a09a2cf7ca50dbad0

pscott:
> Looks good, need to test it
> 
> but looks like https://schnaps.snapshot.box may add turbo to `test.wa0x6e.eth` 😄 we will use different URL on prod?

If we need to disable turbo, we can always create a new payment from the admin [ADMIN](https://github.com/snapshot-labs/schnaps-api/blob/13126288325e01c97e9d1d7182ad2d016bc92587/src/writers.ts#L20) address and specify an old timestamp

wa0x6e:
> > Do we have a dedicated schnaps instance for testnet ?
> 
> No. You can run it locally using `INDEX_TESTNET=true` to have it index `sepolia`.
> 
> Otherwise maybe @bonustrack can deploy something on DO?

Seems like we have https://testnet-schnaps.snapshot.box/

wa0x6e:
> > Looks good, need to test it
> > but looks like https://schnaps.snapshot.box may add turbo to `test.wa0x6e.eth` 😄 we will use different URL on prod?
> 
> If we need to disable turbo, we can always create a new payment from the admin [ADMIN](https://github.com/snapshot-labs/schnaps-api/blob/13126288325e01c97e9d1d7182ad2d016bc92587/src/writers.ts#L20) address and specify an old timestamp

I thought the subscription period was computed from the paid amount, and could not be customized ? e.g if I paid 6k, it will auto extend the subscription by one year

test/.env.test
```diff
@@ -6,3 +6,4 @@ NODE_ENV=test
 RATE_LIMIT_DATABASE_URL=redis://localhost:6379
 RATE_LIMIT_KEYS_PREFIX=snapshot-sequencer-test:
 AUTH_SECRET=5feceb66ffc86f38d952786c6d696c79c2dbc239dd4e91b46729d73a27fb57e9
+NO_SCHNAPS=true
```
wa0x6e, line 9:
```suggestion
```

src/helpers/turbo.ts
```diff
@@ -0,0 +1,87 @@
...
+      // Step 1: Query all the spaces from the schnaps-api
+      const spaces = await getSpacesExpirationDates();
+
+      // Step 2: Update the turbo status of the spaces in the database
+      updateTurboStatuses(spaces);
+
+      // Sleep for a while before running the loop again
+      await snapshot.utils.sleep(RUN_INTERVAL);
+    }
+  }
```
wa0x6e, line 22:
```suggestion
  if (!SCHNAPS_API_URL) return;
  
  while (true) {
    // Step 1: Query all the spaces from the schnaps-api
    const spaces = await getSpacesExpirationDates();

    // Step 2: Update the turbo status of the spaces in the database
    updateTurboStatuses(spaces);

    // Sleep for a while before running the loop again
    await snapshot.utils.sleep(RUN_INTERVAL);
  }
```

pscott:
> > > Looks good, need to test it
> > > but looks like https://schnaps.snapshot.box may add turbo to `test.wa0x6e.eth` 😄 we will use different URL on prod?
> > 
> > 
> > If we need to disable turbo, we can always create a new payment from the admin [ADMIN](https://github.com/snapshot-labs/schnaps-api/blob/13126288325e01c97e9d1d7182ad2d016bc92587/src/writers.ts#L20) address and specify an old timestamp
> 
> I thought the subscription period was computed from the paid amount, and could not be customized ? e.g if I paid 6k, it will auto extend the subscription by one year

This is the case for every user EXCEPT the `ADMIN`. The `ADMIN` is a privileged user that specifies the expiration_timestamp in the IPFS file (trusted). This is done to allow us to give turbo for free, give discounts, etc...

src/helpers/turbo.ts
pscott: Reverted back to if in the condition. Not sure why (probably some weird JS things), but it wasn't working.
Yours: https://github.com/snapshot-labs/snapshot-sequencer/pull/523/commits/80636cfd5f45dd653da1c715d2d92db714b458d9

Revert: https://github.com/snapshot-labs/snapshot-sequencer/pull/523/commits/b169e07da67c6a30751c22041f93a62c9a9c5c9b

wa0x6e: Just tried, and it works. What issue are you seeing ?

a simple `console.log` inside the loop is printing

pscott: I don't know, `yarn test` wasn't working (https://github.com/snapshot-labs/snapshot-sequencer/actions/runs/14125500679/job/39573608892) anymore so I guessed it had broken something. Since it's working in this version I didn't investigate further

wa0x6e: Fixed it https://github.com/snapshot-labs/snapshot-sequencer/actions/runs/14125500679/job/40103852183

Tests sometimes need to be retried, as some of them depends on external services, and can fail

pscott: failing again : https://github.com/snapshot-labs/snapshot-sequencer/actions/runs/14310569196/job/40104292563

wa0x6e (commented):
tAck, but may need one more confirmation as I pushed a few changes

wa0x6e (approved):

src/helpers/turbo.ts
```diff
@@ -0,0 +1,100 @@
...
+    const response = await fetchWithKeepAlive(SCHNAPS_API_URL, {
+      timeout: 5e3,
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ query })
+    });
+
+    const data = await response.json();
```
ChaituVR, line 88: Not too important (for now), but if there are thousands of turbo spaces, will it return all of them without any limits?

src/helpers/turbo.ts
```diff
@@ -0,0 +1,100 @@
...
+// Periodically sync the turbo status of spaces with the schnaps-api
+export async function trackTurboStatuses() {
+  if (!SCHNAPS_API_URL) return;
+
+  while (true) {
+    // Step 1: Query all the spaces from the schnaps-api
+    let spaces = await getSpacesExpirationDates();
+
+    spaces = spaces
+      .filter(space => space.id.startsWith(NETWORK_PREFIX))
```
ChaituVR, line 24: Schnaps api always return `[ { id: 'test.wa0x6e.eth', turbo_expiration: 0 } ]` without prefix `s:` 😅

wa0x6e: > Schnaps api always return `[ { id: 'test.wa0x6e.eth', turbo_expiration: 0 } ]` without prefix `s:` 😅

it does, but only for recent spaces. can not edit old data where spaces were passed without network prefix

src/helpers/turbo.ts
bonustrack: Yes, once we have thousand of Turbo spaces (which should give us a bit of time) we can change that logic.

wa0x6e: @pscott what's the pagination default and max limit?

src/helpers/turbo.ts
```diff
@@ -0,0 +1,100 @@
...
+import db from './mysql';
+import { fetchWithKeepAlive } from './utils';
+
+type Space = {
+  id: string;
+  turbo_expiration: number;
+};
+
+const SCHNAPS_API_URL = process.env.SCHNAPS_API_URL;
+const NETWORK_PREFIX = process.env.NETWORK === '1' ? 's:' : 's-tn:';
```
ChaituVR, line 12: `NETWORK` is either `mainnet` or `testnet`

ChaituVR:
```suggestion
const NETWORK_PREFIX = process.env.NETWORK === 'mainnet' ? 's:' : 's-tn:';
```

wa0x6e:
PR updated to only update `turbo_expiration` field, the `turbo` field is redundant and can be inferred from the expiration date

src/helpers/turbo.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (spaces.length === 0) return;
+
+  // Sync `turbo_expiration` for each space
+  const query = `
+    UPDATE spaces
+    SET
+      turbo_expiration = CASE id
+        ${spaces.map(() => `WHEN ? THEN ?`).join(' ')}
+      END
+    WHERE id IN (${spaces.map(() => '?').join(', ')});
```
bonustrack, line 47: Can't we just use this?
```suggestion
    WHERE id IN (?);
```

src/helpers/turbo.ts
wa0x6e: I've added a `TODO`, we can address this in a few months, when we will reach this limit

ChaituVR (approved):
tAck

PR #527, opened 2025-04-02
feat: use correct network when fetching shib space controller

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/523

Continuation of https://github.com/snapshot-labs/snapshot.js/pull/1138 to use the correct network for space id using .shib names

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -247,3 +247,16 @@ async function updateWalletConnectWhitelist(
 
   return true;
 }
+
+export function getSpaceController(space: string) {
+  const shibariumNetworkMap = {
+    '1': '109',
+    '11155111': '157'
+  };
+
+  return snapshot.utils.getSpaceController(
+    space,
+    space.endsWith('.shib') ? shibariumNetworkMap[DEFAULT_NETWORK] : DEFAULT_NETWORK,
+    { broviderUrl }
+  );
+}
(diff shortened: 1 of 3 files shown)


src/helpers/utils.ts
```diff
@@ -247,3 +247,16 @@ async function updateWalletConnectWhitelist(
...
+    '1': '109',
+    '11155111': '157'
+  };
+
+  return snapshot.utils.getSpaceController(
+    space,
+    space.endsWith('.shib') ? shibariumNetworkMap[DEFAULT_NETWORK] : DEFAULT_NETWORK,
+    { broviderUrl }
+  );
+}
```
ChaituVR, line 262: Better to use `process.env.NETWORK === 'testnet'` to find if it is testnet or not, and to select shibariumNetwork based on that

just to avoid  hardcoding of `11155111` here 


or we can use `SNAPSHOT_ENV` in `src/writer/settings.ts` to pass it here

ChaituVR (changes requested):

src/helpers/utils.ts
ChaituVR: else looking good

ChaituVR (approved):
utAck

src/helpers/utils.ts
```diff
@@ -247,3 +248,16 @@ async function updateWalletConnectWhitelist(
 
   return true;
 }
+
+export function getSpaceController(space: string, network = NETWORK) {
```
ChaituVR, line 252: I think it is better to use `NETWORK` directly inside this function instead of getting same thing from params

But not too important

PR #528, opened 2025-04-03
fix: allow turbo spaces to create proposals with non-premium networks

ChaituVR:
Right now we are not allowing turbo spaces to create proposals with non-premium networks

### How to test
- Create  proposal with turbo space


wa0x6e (approved):
utAck

PR #535, opened 2025-04-24
chore: use pineapple gateway to get JSON on poke

ChaituVR:
Use pineapple gateway on poke

### How to test:
- Attach a network listner
- try going to http://localhost:3001/spaces/testsnapshotspace.eth/poke
- you should see the request is sent to pineapple instead of ipfs.snapshot.box


bonustrack (approved):
utACK

PR #539, opened 2025-04-28
fix: infer turbo status from turbo expiration date

wa0x6e:
Closes https://github.com/snapshot-labs/workflow/issues/548

A space should be considered as `turbo` it the turbo_expiration field is in the future


src/helpers/actions.ts
```diff
@@ -106,7 +106,7 @@ export async function getSpace(id: string, includeDeleted = false, network = DEF
     };
   }
 
-  const query = `SELECT settings, domain, deleted, flagged, verified, turbo, hibernated FROM spaces WHERE id = ? AND deleted in (?) LIMIT 1`;
+  const query = `SELECT settings, domain, deleted, flagged, verified, turbo, turbo_expiration, hibernated FROM spaces WHERE id = ? AND deleted in (?) LIMIT 1`;
```
ChaituVR, line 109: tests fail with `[Error: ER_BAD_FIELD_ERROR: Unknown column 'turbo_expiration' in 'field list']` maybe need to update schema

wa0x6e: updated. Also added an index in the schema file, can you add it to the database ?

test/schema.sql
```diff
@@ -20,6 +21,7 @@ CREATE TABLE spaces (
   INDEX flagged (flagged),
   INDEX hibernated (hibernated),
   INDEX turbo (turbo),
+  INDEX turbo_expiration (turbo_expiration),
```
ChaituVR, line 24: We don't have index on hub's sql and prod db 🤔 we should remove here or add there

ChaituVR (approved):
utAck

test/schema.sql
wa0x6e: Why remove here ? Isn't the goal to have it everywhere ?

ChaituVR: Not sure if we have any advantages of having an index for expiration? if it is important we should add it on hub and prod DB

wa0x6e: It’s to have a faster search for turbo spaces, as we will remove the `turbo` column soon, which is indexed

ChaituVR: ok lets add everywhere then, will create a PR?

ChaituVR: > It’s to have a faster search for turbo spaces, as we will remove the turbo column soon, which is indexed

Do you expect that we will have sorting or queries with turbo_expiration in future? 🤔  don't see anything important with current implementation of `turbo`

wa0x6e: filtering by turbo is expected yes

PR #541, opened 2025-05-21
feat: add delay before finalizing proposal scores

wa0x6e:
This PR fix the issue where a last minute vote just before proposal closing is not taken into account when computing the proposal final score, due to concurrency issue

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -13,6 +13,7 @@ import { name, version } from '../package.json';
 
 const router = express.Router();
 const SNAPSHOT_ENV = process.env.NETWORK || 'testnet';
+const SCORE_FINALIZATION_DELAY = 60000;
 
 const maintenanceMsg = 'update in progress, try later';
 
@@ -47,15 +48,18 @@ router.get('/', (req, res) => {
 });
 
 router.get('/scores/:proposalId', async (req, res) => {
-  const { proposalId } = req.params;
-  try {
-    const result = await serve(proposalId, updateProposalAndVotes, [proposalId]);
-    return res.json({ result });
-  } catch (e) {
-    capture(e);
-    log.warn(`[api] updateProposalAndVotes() failed ${proposalId}, ${JSON.stringify(e)}`);
-    return res.json({ error: 'failed', message: e });
-  }
+  // Delay the score finalization to allow some time for last minute votes to complete
+  setTimeout(async () => {
+    const { proposalId } = req.params;
+    try {
+      await serve(proposalId, updateProposalAndVotes, [proposalId]);
+    } catch (e) {
+      capture(e);
+      log.warn(`[api] updateProposalAndVotes() failed ${proposalId}, ${JSON.stringify(e)}`);
+    }
+  }, SCORE_FINALIZATION_DELAY);
+
+  return res.json({ result: 'pending' });
 });
 
 router.get('/spaces/:key/poke', async (req, res) => {
--- a/src/api.ts
+++ b/src/api.ts
@@ -13,7 +13,6 @@ import { name, version } from '../package.json';
 
 const router = express.Router();
 const SNAPSHOT_ENV = process.env.NETWORK || 'testnet';
-const SCORE_FINALIZATION_DELAY = 60000;
 
 const maintenanceMsg = 'update in progress, try later';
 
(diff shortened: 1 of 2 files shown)


src/api.ts
```diff
@@ -47,15 +48,18 @@ router.get('/', (req, res) => {
...
+    const { proposalId } = req.params;
+    try {
+      await serve(proposalId, updateProposalAndVotes, [proposalId]);
+    } catch (e) {
+      capture(e);
+      log.warn(`[api] updateProposalAndVotes() failed ${proposalId}, ${JSON.stringify(e)}`);
+    }
+  }, SCORE_FINALIZATION_DELAY);
+
+  return res.json({ result: 'pending' });
```
ChaituVR, line 62: I think we should do it inside `updateProposalAndVotes` based on the proposal end time, here we are adding wait time on all requests.

ChaituVR: better to use `snapshot.utils.sleep`?

wa0x6e: Moved the delay to inside `updateProposalAndVotes`

Now, any calls to this function will have a delay, and will process at the earliest, 1 min after the proposal end.

The main API call has been moved back to an sync call, and the call from webhook may taken up to 1min to respond. Upside is that the response is back to a boolean, and will not break UI calls

src/scores.ts
```diff
@@ -108,8 +109,21 @@ export async function updateProposalAndVotes(proposalId: string, force = false)
...
+
+  // Delay computation of final scores, to allow time for last minute votes to finish
+  // up to 1 minute after the end of the proposal
+  if (proposal.end <= ts && proposal.scores_state !== 'final') {
+    const secondsSinceEnd = ts - proposal.end;
+    await snapshot.utils.sleep(
+      Math.min(
+        Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0),
+        FINALIZE_SCORE_SECONDS_DELAY
+      ) * 1000
```
ChaituVR, line 122: `Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0)` can never be bigger than `FINALIZE_SCORE_SECONDS_DELAY` right? 🤔 we can remove `Math.min` here and use `Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0)`

wa0x6e: `FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd` can be negative, in the event the proposal has closed for more than 60 seconds

ChaituVR: No i am talking about `Math.min` not `max`, max is fine 😅 

`Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0)` will always be `<= FINALIZE_SCORE_SECONDS_DELAY` (since we are removing something from it) and `>= 0`, so no benefit of using `Math.min` here, It will never pick 
```suggestion
      Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0) * 1000
```

wa0x6e: You’re right

wa0x6e:
@claude fix the lint issue

src/scores.ts
```diff
@@ -108,8 +109,18 @@ export async function updateProposalAndVotes(proposalId: string, force = false)
...
-  // Ignore score calculation if proposal have more than 100k votes and scores_updated greater than 5 minute
   const ts = Number((Date.now() / 1e3).toFixed());
+
+  // Delay computation of final scores, to allow time for last minute votes to finish
+  // up to 1 minute after the end of the proposal
+  if (proposal.end <= ts && proposal.scores_state !== 'final') {
+    const secondsSinceEnd = ts - proposal.end;
+    await snapshot.utils.sleep(
+      Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0) * 1000
+    );
```
ChaituVR, line 120:
```suggestion
    await snapshot.utils.sleep(Math.max(FINALIZE_SCORE_SECONDS_DELAY - secondsSinceEnd, 0) * 1000);
```

ChaituVR (approved):
utAck

PR #547, opened 2025-06-22
fix: add split-delegation to list of override strategies

wa0x6e:
Add `split-delegation` to the list of overriding strategies


ChaituVR (approved):
utAck

PR #550, opened 2025-06-30
feat: allow starknet aliases to flag proposals

wa0x6e:
Following https://github.com/snapshot-labs/snapshot.js/pull/1164

This PR will allow starknet aliases to flag proposals, when it's a space admin or moderator


ChaituVR (approved):
utAck

PR #554, opened 2025-07-01
fix: implement error handling in moderation data loading loop

bigint:
Replaced the recursive moderation refresh call with a loop that handles errors and sleeps between iterations

### DIFF
--- a/src/helpers/moderation.ts
+++ b/src/helpers/moderation.ts
@@ -42,10 +42,15 @@ export function setData(result?: Record<string, string[]>) {
 }
 
 export default async function run() {
-  setData(await loadModerationData());
+  while (true) {
+    try {
+      setData(await loadModerationData());
+    } catch (e: any) {
+      capture(e);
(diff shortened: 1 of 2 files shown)


src/helpers/moderation.ts
```diff
@@ -42,10 +42,15 @@ export function setData(result?: Record<string, string[]>) {
 }
 
 export default async function run() {
-  setData(await loadModerationData());
+  while (true) {
+    try {
+      setData(await loadModerationData());
+    } catch (e: any) {
+      capture(e);
+    }
```
ChaituVR, line 50: Already have a trycatch inside `loadModerationData` which capture the error , can similify to
```ts
export default async function run() {
  while (true) {
    setData(await loadModerationData());
    await snapshot.utils.sleep(20e3);
  }
}
```

ChaituVR (changes requested):
utAck

PR #560, opened 2025-07-16
feat: prevent spaces from setting themselves as parent or child

ChaituVR:
Related to https://github.com/snapshot-labs/workflow/issues/605

## Summary

This PR adds validation to prevent spaces from creating circular references in the space hierarchy by setting themselves as their own parent or including themselves in their children array.

## Testing
- All existing tests continue to pass
- New tests validate both rejection and acceptance scenarios


wa0x6e (approved):
utAck

PR #562, opened 2025-07-21
feat: add support for sonic domain name

wa0x6e:
Add support for sonic domain name

The SDK is already handling the domain name correctly, this PR updates the code to send the correct network to the `getSpaceController` function when a .sonic domain is detected

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -259,7 +259,7 @@ export function getSpaceController(space: string, network = NETWORK) {
       testnet: '157'
     },
     sonic: {
-      mainnet: '147'
+      mainnet: '146'
     }
   };
   const networkId = tldMapping[tld]?.[network] ?? DEFAULT_NETWORK;


bonustrack (approved):
utACK

src/helpers/utils.ts
```diff
@@ -252,14 +252,17 @@ async function updateWalletConnectWhitelist(
 }
 
 export function getSpaceController(space: string, network = NETWORK) {
-  const shibariumNetworkMap = {
-    mainnet: '109',
-    testnet: '157'
+  const tld = space.split('.').slice(-1)[0];
```
ChaituVR, line 255: Not too important, but this can cause trouble in future if some network in `getSpaceController` doesn't check `endsWith` (we are checking in `getShibariumNameOwner` and `getUDNameOwner` but not `getEnsSpaceController`)


Because 
`"sonic".split('.').slice(-1)[0];` will return `sonic`

It is fine now, but may cause confusion it future because the tldmapping doesn't contain any `.`  but not on snapshot.js https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L48-L57

wa0x6e: it will then default to null, and fallback to `DEFAULT_NETWORK`

ChaituVR: hmm no `sonic` as space name will still go to sonic tld

PR #567, opened 2025-08-03
feat: set proposal vp value on proposal creation

wa0x6e:
This PR will fetch the proposal VP value from the overlord service.

### TODO

- [ ] Set value on proposal creation if possible
- [ ] Monitor webhook to set value on proposal start 
- [ ] Add retry
- [ ] Set different error code on error

### DIFF
--- a/src/helpers/vpValue.ts
+++ b/src/helpers/vpValue.ts
@@ -1,0 +1,58 @@
+import hubDB from './mysql';
+import { fetchWithKeepAlive } from './utils';
+
+type Proposal = {
+  id: string;
+  network: string;
+  strategies: any[];
+  start: number;
+};
+
+const OVERLORD_URL = 'https://overlord.snapshot.org';
+const CB_LAST = 5;
+const CB_ERROR = 0;
+
+export async function setProposalVpValue(proposal: Proposal) {
+  if (proposal.start > Date.now()) {
+    return;
+  }
+
+  try {
+    const vpValue = await getVpValue(proposal);
+
+    await hubDB.queryAsync('UPDATE proposals SET vp_value = ?, cb = ? WHERE id = ? LIMIT 1', [
+      vpValue,
+      CB_LAST,
+      proposal.id
+    ]);
+  } catch {
+    await hubDB.queryAsync('UPDATE proposals SET cb = ? WHERE id = ? LIMIT 1', [
+      CB_ERROR,
+      proposal.id
+    ]);
+    return;
+  }
+}
+
+async function getVpValue(proposal: Proposal) {
+  const init = {
+    method: 'POST',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({
+      jsonrpc: '2.0',
+      method: 'get_vp_value_by_strategy',
+      params: {
+        network: proposal.network,
+        strategies: proposal.strategies,
+        snapshot: proposal.start
+      },
+      id: Math.random().toString(36).substring(7)
+    })
+  };
+  const res = await fetchWithKeepAlive(OVERLORD_URL, init);
+  const { result } = await res.json();
+  return result;
+}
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -19,4 +19,5 @@ STARKNET_RPC_URL= # optional
 AUTH_SECRET=1dfd84a695705665668c260222ded178d1f1d62d251d7bee8148428dac6d0487 # optional
 WALLETCONNECT_PROJECT_ID=e6454bd61aba40b786e866a69bd4c5c6
 REOWN_SECRET= # optional
+OVERLORD_URL=https://overlord.snapshot.org
```
bonustrack, line 22:
```suggestion
OVERLORD_URL=https://overlord.snapshot.box
```

src/helpers/entityValue.ts
```diff
@@ -1,16 +1,39 @@
...
+  network: string;
+  strategies: any[];
+  start: number;
 };
 
+const OVERLORD_URL = process.env.OVERLORD_URL ?? 'https://overlord.snapshot.org';
+// Round strategy values to 9 decimal places
+const STRATEGIES_VALUE_PRECISION = 9;
+
+export async function getStrategiesValue(proposal: Proposal): Promise<number[]> {
```
bonustrack, line 17: Ideally we should keep same naming everywhere, here I would just call it  `getVpValueByStrategy`

wa0x6e:
Code and PR description updated to the new spec

src/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const CB = {
+  INELIGIBLE: -1,
+  PENDING_SYNC: -10, // waiting for value from overlord
+  PENDING_COMPUTE: -15, // value from overlord set, waiting for local computation or refresh
+  PENDING_CLOSE: -20 // value from overlord set, waiting for proposal close for final computation
+};
+
+export const CURRENT_CB = parseInt(process.env.LAST_CB ?? '1');
```
bonustrack, line 8: I was thinking actually the final cb could be always `1`. In case we want to resync we can simply update the entries we want resync with a PENDING_SYNC cb, we dont need to change the final cb necessary.

src/helpers/proposalStrategiesValue.ts
```diff
@@ -0,0 +1,70 @@
...
+async function refreshProposalsVpValues(proposals: Proposal[]) {
+  const query: string[] = [];
+  const params: any[] = [];
+
+  for (const proposal of proposals) {
+    await buildQuery(proposal, query, params);
+  }
+
+  if (query.length) {
+    await db.queryAsync(query.join(';'), params);
```
bonustrack, line 37: This is confusing, I would expect a function like `buildQuery` to return the `query` var instead of instantiate the variable in the `refreshProposalsVpValues` function to let it updated in `buildQuery`? It feel like with this approach depending on how we call the function it could conflict

bonustrack (commented):
Added some suggestions otherwise it looks good

src/helpers/proposalStrategiesValue.ts
```diff
@@ -0,0 +1,70 @@
...
+      console.log(e);
+    }
+  }
+
+  if (query.length) {
+    await db.queryAsync(query.join(';'), params);
+  }
+}
+
+export default async function run() {
```
bonustrack, line 56:
```suggestion
async function run() {
```

bonustrack (approved):
utACK

src/helpers/proposalStrategiesValue.ts
```diff
@@ -0,0 +1,82 @@
...
+    params.push(JSON.stringify(result.values), result.cb, result.proposal.id);
+  }
+
+  if (query.length) {
+    await db.queryAsync(query.join(';'), params);
+  }
+}
+
+export default async function run() {
+  while (true) {
```
wa0x6e, line 59:
```suggestion
  while (true) {
    const proposals = await getProposals();

    if (proposals.length) {
      await refreshVpByStrategy(proposals);
    }

    if (proposals.length < BATCH_SIZE) {
      await snapshot.utils.sleep(REFRESH_INTERVAL);
    }
  }
```

PR #571, opened 2025-08-06
feat: add vp value to leaderboard

wa0x6e:
## Summary

Adds `vp_value` tracking to the leaderboard table and defers vote deletion to an async background script. Leaderboard updates use idempotent SUM/COUNT queries from the votes table instead of fragile delta-based increments, preventing drift from re-votes, race conditions, or missed updates.

Background workers (`votesVpValue`, `scores`, `shutter`) now guard against overwriting `cb=PENDING_DELETE` set by `delete-proposal`, preventing a race condition where votes marked for deletion could be orphaned.

## Changes

- ✨ Add `vp_value` column to the `leaderboard` table
- ✨ Add `PENDING_DELETE` constant (`cb = -3`) for soft-deleting votes
- ✨ Add `deleteProposalVotes` async script to process pending vote deletions in batches
- ♻️ `delete-proposal` writer: mark votes as `PENDING_DELETE` instead of deleting inline
- ♻️ `votesVpValue`: refresh leaderboard `vp_value` using `SUM(vp_value)` from votes table (idempotent)
- ♻️ `deleteProposalVotes`: refresh leaderboard `vote_count` and `vp_value` using `COUNT(*)`/`SUM()` via single JOIN subquery (idempotent)
- 🐛 `votesVpValue`: guard vote `cb` update with `AND cb = PENDING_COMPUTE` to prevent overwriting `PENDING_DELETE`
- 🐛 `scores`: guard vote update with `AND cb != PENDING_DELETE` to prevent overwriting deletion state
- 🐛 `shutter`: skip `PENDING_DELETE` votes during decryption to avoid wasted work and inconsistency

## Benchmark

Tested against production database (68M+ votes). All queries use the PRIMARY KEY `(voter, space, proposal)` index.

| Query | Time |
|---|---|
| `SUM(vp_value) WHERE voter+space` (single pair) | ~500ms |
| Batch leaderboard refresh (50 pairs) | 1,123ms |
| Batch leaderboard refresh (200 pairs) | 797ms |
| Batch leaderboard refresh (500 pairs) | 1,355ms |

~500ms is mostly network latency to PlanetScale. The query itself is near-instant (PK index lookup).

## Test plan

- [ ] Verify `vp_value` column is added to leaderboard table
- [ ] Verify votes are marked `cb = -3` on proposal deletion instead of being deleted inline
- [ ] Verify `deleteProposalVotes` script picks up and deletes pending votes
- [ ] Verify leaderboard `vp_value` is refreshed correctly after vote vp_value computation
- [ ] Verify leaderboard `vote_count` and `vp_value` are correct after vote deletion
- [ ] Verify background workers do not overwrite `PENDING_DELETE` on votes mid-deletion

### DIFF
--- a/src/helpers/deleteProposalVotes.ts
+++ b/src/helpers/deleteProposalVotes.ts
@@ -37,30 +37,27 @@ async function processVotes(votes: Vote[]) {
   query.push('DELETE FROM votes WHERE id IN (?)');
   params.push(votes.map(v => v.id));
 
-  await db.queryAsync(query.join(';'), params);
-
-  // Refresh leaderboard from remaining votes (idempotent, single batched query)
+  // Refresh leaderboard from remaining votes (idempotent)
   const pairs = new Set(votes.map(v => `${v.voter}:${v.space}`));
-  if (!pairs.size) return;
-
-  const pairPlaceholders = Array.from(pairs)
-    .map(() => '(?, ?)')
-    .join(', ');
-  const pairParams: string[] = [];
-  for (const pair of pairs) {
-    const [voter, space] = pair.split(':');
-    pairParams.push(voter, space);
+  if (pairs.size) {
+    const pairPlaceholders = Array.from(pairs)
+      .map(() => '(?, ?)')
+      .join(', ');
+    query.push(
+      `UPDATE leaderboard l
+        SET vote_count = (SELECT COUNT(*) FROM votes v WHERE v.voter = l.user AND v.space = l.space),
+            vp_value = COALESCE((
+              SELECT SUM(v.vp_value) FROM votes v WHERE v.voter = l.user AND v.space = l.space
+            ), 0)
+        WHERE (l.user, l.space) IN (${pairPlaceholders})`
+    );
+    for (const pair of pairs) {
+      const [voter, space] = pair.split(':');
+      params.push(voter, space);
+    }
   }
 
-  await db.queryAsync(
-    `UPDATE leaderboard l
-      SET vote_count = (SELECT COUNT(*) FROM votes v WHERE v.voter = l.user AND v.space = l.space),
-          vp_value = COALESCE((
-            SELECT SUM(v.vp_value) FROM votes v WHERE v.voter = l.user AND v.space = l.space
-          ), 0)
-      WHERE (l.user, l.space) IN (${pairPlaceholders})`,
-    pairParams
-  );
+  await db.queryAsync(query.join(';'), params);
 }
 
 export default async function run() {
--- a/src/helpers/deleteProposalVotes.ts
+++ b/src/helpers/deleteProposalVotes.ts
@@ -42,7 +42,9 @@ async function processVotes(votes: Vote[]) {
   // Refresh leaderboard from remaining votes (idempotent, single batched query)
   const pairs = new Set(votes.map(v => `${v.voter}:${v.space}`));
   if (pairs.size > 0) {
-    const pairPlaceholders = Array.from(pairs).map(() => '(?, ?)').join(', ');
+    const pairPlaceholders = Array.from(pairs)
+      .map(() => '(?, ?)')
+      .join(', ');
     const pairParams: string[] = [];
     for (const pair of pairs) {
       const [voter, space] = pair.split(':');
--- a/src/helpers/deleteProposalVotes.ts
+++ b/src/helpers/deleteProposalVotes.ts
@@ -34,8 +34,11 @@ async function processVotes(votes: Vote[]) {
   }
 
   // Delete votes
-  query.push('DELETE FROM votes WHERE id IN (?)');
-  params.push(votes.map(v => v.id));
+  query.push('DELETE FROM votes WHERE id IN (?) AND cb = ?');
+  params.push(
+    votes.map(v => v.id),
+    CB.PENDING_DELETE
+  );
 
   // Refresh leaderboard from remaining votes (idempotent)
   const pairs = new Set(votes.map(v => `${v.voter}:${v.space}`));
--- a/src/helpers/deleteProposalVotes.ts
+++ b/src/helpers/deleteProposalVotes.ts
@@ -1,0 +1,74 @@
+import snapshot from '@snapshot-labs/snapshot.js';
+import log from './log';
+import db from './mysql';
+import { CB } from '../constants';
+
+type Vote = {
+  id: string;
+  voter: string;
+  space: string;
+  vp_value: number;
+  vp_state: string;
+};
+
+const REFRESH_INTERVAL = 60 * 1000;
+const BATCH_SIZE = 100;
+
+async function getVotesPendingDeletion(): Promise<Vote[]> {
+  return db.queryAsync(
+    `SELECT id, voter, space, vp_value, vp_state FROM votes WHERE cb = ? LIMIT ?`,
+    [CB.PENDING_DELETE, BATCH_SIZE]
+  );
+}
+
+async function processVotes(votes: Vote[]) {
+  const query: string[] = [];
+  const params: (string | number | string[])[] = [];
+
+  // Update spaces vote_count
+  const grouped = new Map<string, Vote[]>();
+  for (const vote of votes) {
+    if (!grouped.has(vote.space)) {
+      grouped.set(vote.space, []);
+    }
+    grouped.get(vote.space)!.push(vote);
+  }
+
+  for (const [space, spaceVotes] of grouped) {
+    query.push('UPDATE spaces SET vote_count = GREATEST(vote_count - ?, 0) WHERE id = ?');
+    params.push(spaceVotes.length, space);
+  }
+
+  // Update leaderboard vote_count and vp_value
+  for (const vote of votes) {
+    query.push(
+      `UPDATE leaderboard
+        SET vote_count = GREATEST(vote_count - 1, 0),
+            vp_value = GREATEST(vp_value - ?, 0)
+        WHERE user = ? AND space = ?`
+    );
+    params.push(vote.vp_state === 'final' ? vote.vp_value : 0, vote.voter, vote.space);
+  }
+
+  // Delete votes
+  query.push('DELETE FROM votes WHERE id IN (?)');
+  params.push(votes.map(v => v.id));
+
+  await db.queryAsync(query.join(';'), params);
+}
+
+export default async function run() {
+  while (true) {
+    const votes = await getVotesPendingDeletion();
+
+    if (votes.length) {
+      log.info(`[deleteProposalVotes] ${votes.length} votes to delete`);
+      await processVotes(votes);
+    }
+
+    if (votes.length < BATCH_SIZE) {
+      log.info('[deleteProposalVotes] sleeping');
+      await snapshot.utils.sleep(REFRESH_INTERVAL);
+    }
+  }
+}
--- a/src/helpers/deleteProposalVotes.ts
+++ b/src/helpers/deleteProposalVotes.ts
@@ -24,39 +24,41 @@ async function processVotes(votes: Vote[]) {
   const params: (string | number | string[])[] = [];
 
   // Update spaces vote_count
-  const grouped = new Map<string, Vote[]>();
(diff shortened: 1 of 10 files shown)


src/writer/delete-proposal.ts
```diff
@@ -53,4 +55,21 @@ export async function action(body): Promise<void> {
...
+        .map(
+          () =>
+            `UPDATE leaderboard SET vp_value = GREATEST(vp_value - ?, 0) WHERE user = ? AND space = ?;`
+        )
+        .join('\n    ');
+
+      const vpParams = batch.flatMap(voter => [voter.vp_value, voter.voter, msg.space]);
+
+      await db.queryAsync(vpQueries, vpParams);
+    }
```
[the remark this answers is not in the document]
ChaituVR: I think maybe it is better if we maintain a db or table or files to keep the deleted proposals/votes (like an archive) the current approach can cause wrong values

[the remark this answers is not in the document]
wa0x6e: Another safer way to handle that would be use an async script, with flow like this:

- on proposal delete, keep old code from delete-proposal.ts, but we will update the CB to `PENDING_COMPUTE`, and not delete any votes yet.

An async script will pickup the `proposal.deleted = 1 AND proposals.cb = pending_compute`, and process them:

- going through all the votes in batch, and send queries like `DELETE FROM votes ...; UPDATE leaderboard;` in the same query (so we don't lose data)
- after all votes and leaderboard are updated, switch the proposal CB to FINAL

In case of any errors (server restart), the async script can pickup where it crashes)

src/helpers/votesVpValue.ts
```diff
@@ -110,6 +127,17 @@ async function refreshVotesVpValues(data: Datum[]) {
...
+      // Leaderboard update only for final votes
+      // to avoid vp fluctuations with overriding strategies
+      // Use delta (new - old) to handle re-votes correctly
+      if (validatedDatum.vpState === 'final') {
+        leaderboardUpdates.push({
+          value: value - datum.vpValue,
+          voter: validatedDatum.voter,
+          space: validatedDatum.space
+        });
+      }
```
ChaituVR: Looks like when proposal is open we are setting vp_value in votes table but not on leaderboard, this is because vpState can be `pending` for proposals using override strategies

when proposal is closed, on votes_table we store the `value` but on leaderboard it is subtracting `value - datum.vpValue` which will make it `0` always

ChaituVR (changes requested):

src/helpers/votesVpValue.ts
wa0x6e: Yes, we handle this case implicitly, assuming that the `state` flow will prevent weird issues.

Do you thing making things more explicit will help, by adding a guard to decrease only when vpState is `final`, and also only when the `vpValue` is not-null (to remove unecessary UPDATE) ?

src/helpers/votesVpValue.ts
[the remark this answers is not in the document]
wa0x6e: This is for the votes table.

We're setting the vp_value, even on pending and overriding strategies for the votes, only leaderboard will update only on `final`.

Unless we want to handle all cases together:

- update votes and leaderboard only on final
- always update votes and leaderboard, even when not final (with the vp_value delta introduced from last commit, we now know the difference with the previous votes, so we could handle leaderboard update on votes update), but this could flaky. We're fine with only one server instance, but this async script could double process when the same votes is picked up by multiple instances.

src/helpers/votesVpValue.ts
ChaituVR: Right now it doesn't add up, we may always end up with `0` no? 🤔

src/writer/delete-proposal.ts
```diff
@@ -21,36 +22,19 @@ export async function verify(body): Promise<any> {
...
+    UPDATE votes SET cb = ? WHERE proposal = ?;
     UPDATE leaderboard
       SET proposal_count = GREATEST(proposal_count - 1, 0)
       WHERE user = ? AND space = ?
       LIMIT 1;
     UPDATE spaces
-      SET proposal_count = GREATEST(proposal_count - 1, 0), vote_count = GREATEST(vote_count - ?, 0)
+      SET proposal_count = GREATEST(proposal_count - 1, 0)
       WHERE id = ?;
   `;
```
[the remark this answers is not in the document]
ChaituVR: hmm yea we can ignore i think

src/helpers/votesVpValue.ts
wa0x6e: What do you mean does not add up ? 

This computation may result in value = 0, we could add a guard to no update when old value same as new value, to save a few queries. It was left out to keep the changes minimal

ChaituVR: Overriding strategies is not a edge case right? not sure if its good idea to ignore this case

ChaituVR: I mean with this leaderboard never gets the value at all, for all proposals using override strategies

wa0x6e: They will, when the vpState switch to final

ChaituVR: AI explanation:

The leaderboard block runs when vpState switches to final — that's not the issue.

The issue is what value it adds. The delta is `value - datum.vpValue`:

Run 1 (pending): writes `vp_value = 100` to votes table, skips leaderboard ✅
Run 2 (final): `value = 100`, `datum.vpValue = 100 `(from run 1) → `delta = 100 - 100 = 0`
Leaderboard gets +0
The code enters the if (vpState === 'final') block, yes — but with a delta of 0. So it fires an UPDATE that adds nothing.

wa0x6e: Ah, I see

Maybe should always update on all votes, not only when final, or not use increase/decrease, and use `sum` from sql to get the exact total vp (will avoid double processing in case we have multiple instances running)

ChaituVR: `sum` is good i think

src/helpers/deleteProposalVotes.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+  for (const [space, count] of grouped) {
+    query.push('UPDATE spaces SET vote_count = GREATEST(vote_count - ?, 0) WHERE id = ?');
+    params.push(count, space);
+  }
+
+  // Delete votes
+  query.push('DELETE FROM votes WHERE id IN (?)');
+  params.push(votes.map(v => v.id));
+
```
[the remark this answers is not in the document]
wa0x6e: Will live with that edge case, as a `select count()` in the votes table is very slow, without `voter` in the where filter

ChaituVR:
Claude says this:  (But i didn't verify code myself but it sounds reasonable)

## **The Orphaned Vote Bug — Step by Step**

Setup: Alice voted on Proposal P1 in space aave.eth. Her vote has cb = PENDING_COMPUTE, vp_value = 0.

**Step 0**: votesVpValue wakes up, starts its cycle.
**Step 1**: getPendingVotes() runs:

```SQL
SELECT id, voter, space, proposal FROM votes WHERE cb = -1 LIMIT 500
Returns: { id: 'vote_abc', voter: 'alice', space: 'aave.eth', proposal: 'P1' }
```

**Steps 2**: getProposalVpValues() runs:

```SQL
SELECT id, cb, vp_value_by_strategy FROM proposals WHERE cb IN (-1,-2,1,-10) AND votes > 0
Returns: { id: 'P1', vpValueByStrategy: [0.5, 1.2] } — proposal exists, all good.
```

**Step 3:** A moderator deletes Proposal P1. delete-proposal.ts runs atomically:

```SQL
DELETE FROM proposals WHERE id = 'P1';
UPDATE votes SET cb = -3 WHERE proposal = 'P1';  -- PENDING_DELETE
UPDATE leaderboard SET proposal_count = GREATEST(proposal_count - 1, 0) WHERE user = 'mod' AND space = 'aave.eth';
UPDATE spaces SET proposal_count = GREATEST(proposal_count - 1, 0) WHERE id = 'aave.eth';
Alice's vote is now cb = -3 (PENDING_DELETE) in the database.
```

**Step 4**: votesVpValue doesn't know any of this happened. It still has the old data from Time 1-2. It computes vp_value = 100 and runs refreshVotesVpValues():

```SQL
UPDATE votes SET vp_value = 100, cb = 1 WHERE id IN ('vote_abc');
UPDATE leaderboard l SET vp_value = COALESCE((
  SELECT SUM(v.vp_value) FROM votes v WHERE v.voter = l.user AND v.space = l.space
), 0) WHERE (l.user, l.space) IN (('alice', 'aave.eth'));
```

### This overwrites cb = -3 (PENDING_DELETE) with cb = 1 (FINAL).

wa0x6e:
> Claude says this: (But i didn't verify code myself but it sounds reasonable)
> 
> ## **The Orphaned Vote Bug — Step by Step**
> Setup: Alice voted on Proposal P1 in space aave.eth. Her vote has cb = PENDING_COMPUTE, vp_value = 0.
> 
> **Step 0**: votesVpValue wakes up, starts its cycle. **Step 1**: getPendingVotes() runs:
> 
> ```sql
> SELECT id, voter, space, proposal FROM votes WHERE cb = -1 LIMIT 500
> Returns: { id: 'vote_abc', voter: 'alice', space: 'aave.eth', proposal: 'P1' }
> ```
> 
> **Steps 2**: getProposalVpValues() runs:
> 
> ```sql
> SELECT id, cb, vp_value_by_strategy FROM proposals WHERE cb IN (-1,-2,1,-10) AND votes > 0
> Returns: { id: 'P1', vpValueByStrategy: [0.5, 1.2] } — proposal exists, all good.
> ```
> 
> **Step 3:** A moderator deletes Proposal P1. delete-proposal.ts runs atomically:
> 
> ```sql
> DELETE FROM proposals WHERE id = 'P1';
> UPDATE votes SET cb = -3 WHERE proposal = 'P1';  -- PENDING_DELETE
> UPDATE leaderboard SET proposal_count = GREATEST(proposal_count - 1, 0) WHERE user = 'mod' AND space = 'aave.eth';
> UPDATE spaces SET proposal_count = GREATEST(proposal_count - 1, 0) WHERE id = 'aave.eth';
> Alice's vote is now cb = -3 (PENDING_DELETE) in the database.
> ```
> 
> **Step 4**: votesVpValue doesn't know any of this happened. It still has the old data from Time 1-2. It computes vp_value = 100 and runs refreshVotesVpValues():
> 
> ```sql
> UPDATE votes SET vp_value = 100, cb = 1 WHERE id IN ('vote_abc');
> UPDATE leaderboard l SET vp_value = COALESCE((
>   SELECT SUM(v.vp_value) FROM votes v WHERE v.voter = l.user AND v.space = l.space
> ), 0) WHERE (l.user, l.space) IN (('alice', 'aave.eth'));
> ```
> 
> ### This overwrites cb = -3 (PENDING_DELETE) with cb = 1 (FINAL).

A valid concern, now fixed

wa0x6e:
@ChaituVR Race conditions should be fixed now

src/helpers/deleteProposalVotes.ts
```diff
@@ -0,0 +1,77 @@
...
+  for (const vote of votes) {
+    grouped.set(vote.space, (grouped.get(vote.space) || 0) + 1);
+  }
+  for (const [space, count] of grouped) {
+    query.push('UPDATE spaces SET vote_count = GREATEST(vote_count - ?, 0) WHERE id = ?');
+    params.push(count, space);
+  }
+
+  // Delete votes
+  query.push('DELETE FROM votes WHERE id IN (?)');
```
ChaituVR: Maybe we can add additional guard  with `CB.PENDING_DELETE`
```suggestion
  query.push('DELETE FROM votes WHERE id IN (?) AND cb = -3');
```

src/helpers/deleteProposalVotes.ts
```diff
@@ -0,0 +1,77 @@
...
+  params.push(votes.map(v => v.id));
```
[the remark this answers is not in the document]
ChaituVR: Like this

ChaituVR (approved):
utAck,  but haven't tested 🙈

PR #573, opened 2025-08-07
chore: remove polygon

wa0x6e:
Polygon have been removed from sx

This PR will removes polling sx for polygon, since the network does not exist anymore, and will fix the failing test.


bonustrack (approved):
utACK

PR #587, opened 2025-09-03
fix: ignore vote vp_value computation when proposal vp_value is empty

wa0x6e:
When the proposal `vp_value_by_strategy` is empty, the vp value computation is throwing an error. 

This is to catch any error when the both arrays size does not match, but we should ignore this when the array is empty, to prevent noise in the error logs.

This PR will throw an error only when the array is not empty, and with size not matching


ChaituVR (approved):
utAck

PR #588, opened 2025-09-03
fix: fix missing conversion from string to array

wa0x6e:
Fix https://snapshot-labs.sentry.io/issues/6854066828/?project=4505606761152512&query=is%3Aunresolved&referrer=issue-stream

This PR fix the missing conversion from string to array type


ChaituVR (approved):
utAck

PR #593, opened 2025-09-29
feat(overlord): add async script to set vote vp value

wa0x6e:
In continuation to https://github.com/snapshot-labs/snapshot-sequencer/pull/567

This PR will add a new async script, which will periodically update votes' `vp_value` when necessary.

## Goal

We want to assign a VP value (usd value equivalent to the vote's voting power) to all eligible votes. (vp value is available only on some strategies, and tokens, irrelevant to this PR)

This is a 2 process step:

1. Getting the unit VP value of the proposal per vp and per strategy from overlord API (so we can infer 1 VP = xxx USD), stored in `proposals.vp_value_by_strategy` column, will be an array where each element is the usd value per strategy (e.g. `[0.3, 1.5, 0, 0]`)
2. Assigning a VP value for each votes, based on the proposal VP value. Essentially, getting the proposal VP value, and multiply by the votes voting power, per strategy. This a simple local computation. (e.g. if the proposal VP value is `[0.3, 1.5, 0, 0]`, and the vote's VP per strategy is `[1, 2, 3, 4]`, we multiply both array to have `[0.3, 3, 0, 0]`, and sum it, to have `3.3` usd)

This PR is about step 2, where we handle how to compute the votes VP value, and the way we compute it (since it can have multiple "state", defined by the `CB` column in the votes table).

The computation logic already exist, and previously, this was computed in live on votes creation and update, this PR is moving the computation to an async script

## Flow

There is an infinite loop to compute the votes vp value async

- the script will fetch all votes with cb `PENDING_COMPUTE` (default value on vote creation).
- if proposal's `vp_value_by_strategy` is not ready, skip and move to next vote
- else, compute a new `vp_value` as well as assign a new CB (`FINAL` when the vp_state is final, `PENDING_FINAL` otherwise) to the vote
- When a new vote with overriding strategy is added, all the votes will be set to `PENDING_COMPUTE`
- Sleep, and loop again

The flow is controlled partly by the CB state:

- `PENDING_COMPUTE` : pending computation by the script
- `PENDING_FINAL` : votes vp value is set, and work is done from this script perspective. A future async script (from an upcoming PR) will pickup and process those votes
- `FINAL` : vp value is set and final
- `INELIGIBLE` : for invalid data

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,6 @@ SCHNAPS_API_URL=https://schnaps.snapshot.box
 RATE_LIMIT_DATABASE_URL= # optional
 RATE_LIMIT_KEYS_PREFIX=snapshot-sequencer: # optional
 BROVIDER_URL=https://rpc.snapshot.org # optional
-STARKNET_RPC_URL= # optional
 # Secret for laser
 AUTH_SECRET=1dfd84a695705665668c260222ded178d1f1d62d251d7bee8148428dac6d0487 # optional
 WALLETCONNECT_PROJECT_ID=e6454bd61aba40b786e866a69bd4c5c6
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@snapshot-labs/pineapple": "^1.2.1",
     "@snapshot-labs/snapshot-metrics": "^1.4.1",
     "@snapshot-labs/snapshot-sentry": "^1.5.5",
-    "@snapshot-labs/snapshot.js": "^0.14.14",
+    "@snapshot-labs/snapshot.js": "^0.14.17",
     "bluebird": "^3.7.2",
     "connection-string": "^1.0.1",
     "cors": "^2.8.5",
--- a/src/helpers/proposalStrategiesValue.ts
+++ b/src/helpers/proposalStrategiesValue.ts
@@ -56,6 +56,7 @@ async function refreshVpByStrategy(proposals: Proposal[]) {
 }
 
 export default async function run() {
+  if (!process.env.OVERLORD_URL) return;
   while (true) {
     const proposals = await getProposals();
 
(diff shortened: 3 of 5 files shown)


src/helpers/votesVpValue.ts
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      const validatedDatum = datumSchema.parse(datum);
+      const value = validatedDatum.vpValueByStrategy.reduce(
+        (sum, value, index) => sum + value * validatedDatum.vpByStrategy[index],
+        0
+      );
+
+      params.push(
+        value,
+        validatedDatum.vpState === 'final' ? CB.FINAL : CB.PENDING_FINAL,
```
bonustrack: How does a vote goes from the state PENDING_FINAL to FINAL?

wa0x6e:
> > if proposal's vp_value_by_strategy is not ready, skip and move to next vote
> 
> Is this enforced in the code? I can't seem to find it.

This is implicitly handled when JOINING the votes with the proposals tables, we're filtering out votes where the proposal's CB is not ready:

https://github.com/snapshot-labs/snapshot-sequencer/pull/593/changes#diff-43eae564b2e00462cd538df9278c84d875d9ca7d2c33fa154ed56fe00dfaa2e4R37

src/helpers/votesVpValue.ts
wa0x6e: When a vote CB is `PENDING_FINAL`, its vpState will be `pending`

When the vpState move from `pending` to `final`, the CB of the vote will be reset to `PENDING_COMPUTE`, from https://github.com/snapshot-labs/snapshot-sequencer/pull/593/changes#diff-5bcad33ffda7ca294c0f9f823153603ed41f2142a3290946e9595bcc85ae5649L187

The script will pick up and process the votes again, and move it to `FINAL`

In summary, it don't go from `PENDING_FINAL` to `FINAL` directly, but `PENDING_FINAL` => `PENDING_COMPUTE` => `FINAL`

bonustrack (approved):
utACK

PR #629, opened 2026-03-14
fix: allow alias renewal by re-submitting existing pair

wa0x6e:
## Summary

Re-submitting an existing (address, alias) pair was failing with a MySQL duplicate key error. This was problematic for bots and agents that use fixed addresses, as they couldn't renew or extend an alias once created.

Now the same pair can be re-submitted at any time (whether active or expired), which resets the `created` timestamp and effectively renews the alias. Uses `ON DUPLICATE KEY UPDATE` to handle this at the database level.

## Changes

- 🔧 Changed `INSERT` to `INSERT ... ON DUPLICATE KEY UPDATE` in `action()` to refresh `id`, `ipfs`, and `created` on re-submission
- 🧪 Added integration tests for alias renewal and created timestamp bump
- 🧹 Fixed describe block name in unit test (`writer/profile` → `writer/alias`)

## Test plan

- [ ] Verify alias unit tests pass (`yarn test:unit`)
- [ ] Verify integration tests pass (`yarn test:integration`)
- [ ] Manually test re-submitting an existing alias pair succeeds and updates the `created` timestamp

### DIFF
--- a/src/writer/alias.ts
+++ b/src/writer/alias.ts
@@ -10,9 +10,10 @@ export async function verify(message): Promise<any> {
     log.warn(`[writer] Wrong alias format ${JSON.stringify(schemaIsValid)}`);
     return Promise.reject('wrong alias format');
   }
-  if (message.from === msg.payload.alias) {
+  if (message.address === msg.payload.alias) {
     return Promise.reject('alias cannot be the same as the address');
   }
+
   return true;
 }
 
@@ -25,5 +26,8 @@ export async function action(message, ipfs, receipt, id): Promise<void> {
     alias: msg.payload.alias,
     created: msg.timestamp
   };
-  await db.queryAsync('INSERT INTO aliases SET ?', params);
+  await db.queryAsync(
+    'INSERT INTO aliases SET ? ON DUPLICATE KEY UPDATE id = VALUES(id), ipfs = VALUES(ipfs), created = VALUES(created)',
+    params
+  );
 }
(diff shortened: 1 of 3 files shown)


src/writer/alias.ts
```diff
@@ -10,9 +10,18 @@ export async function verify(message): Promise<any> {
...
     return Promise.reject('alias cannot be the same as the address');
   }
+
+  const existing = await db.queryAsync(
+    'SELECT 1 FROM aliases WHERE address = ? AND alias = ? LIMIT 1',
+    [message.address, msg.payload.alias]
+  );
+  if (existing.length > 0) {
+    return Promise.reject('alias already exists');
+  }
```
ChaituVR: Actually I think the main problem is that bots will have fixed address (mostly), if we add them as alias and is expired, we won't be able to add alias with that address again (in case the alias expire)? 🤔

wa0x6e: We should ask the bot to regenerate a new address each time. Later, we could pass a timestamp for the expiration, so the bot will store and send it along the alias, and will know beforehand when the key will expire.

Re-using an expired key is not recommended security-wise

bonustrack:
Is this to enforce a single user have only 1 alias? If so it wont work, users might be logged on many devices we need multiple working alias for a single user to be possible.

src/writer/alias.ts
ChaituVR: hmm not sure if thats the case with agents @bonustrack wyt?

the way i assume, its pain to keep updating address of agents, we need a way to extend/refresh the expiration if user want to

wa0x6e:
No, it’s to enforce alias is unique, and not associated to multiple adresses

src/writer/alias.ts
wa0x6e: PR update. Re-read the description, as this change stray significantly from the original design:

- original: duplicate key was already block on sql level, the pr throw a better error
- new: duplicate key is now accepted

ChaituVR (approved):
tAck

PR #630, opened 2026-03-16
fix: ensure alias can only be linked to a single address

wa0x6e:
## Summary

An alias could previously be linked to multiple addresses, allowing different users to claim the same alias. This PR adds a server-side check in `verify()` to reject alias creation when the alias is already linked to a different address, while still allowing the same address to renew its existing alias.

## Changes

- ✅ `verify()`: rejects when alias is linked to a different address (single SQL query with `AND address != ?`)
- ✅ `action()`: uses `ON DUPLICATE KEY UPDATE` for renewals (from master via PR #629)
- ✅ Added `UNIQUE KEY` on `aliases.alias` in test schema
- ✅ Updated fixtures to avoid `UNIQUE KEY` collision
- ✅ Integration test for cross-address alias rejection
- ✅ Unit test with DB mock since `verify()` now queries the DB

## Database migration

Run the following on the production database to enforce uniqueness at the DB level:

```sql
ALTER TABLE aliases ADD UNIQUE KEY alias (alias);
```

## Test plan

- `yarn test:unit` — alias unit tests pass with mocked DB
- `yarn test:integration` — alias integration tests pass, including new cross-address rejection test
- `yarn test:e2e` — existing e2e tests unaffected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage
 # Remove some common IDE working directories
 .idea
 .vscode
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ coverage
 # Remove some common IDE working directories
 .idea
 .vscode
-package-lock.json
--- a/src/writer/alias.ts
+++ b/src/writer/alias.ts
@@ -14,10 +14,11 @@ export async function verify(message): Promise<any> {
     return Promise.reject('alias cannot be the same as the address');
   }
 
-  const [existing] = await db.queryAsync('SELECT address FROM aliases WHERE alias = ? LIMIT 1', [
-    msg.payload.alias
-  ]);
-  if (existing && existing.address.toLowerCase() !== message.address.toLowerCase()) {
+  const [existing] = await db.queryAsync(
+    'SELECT address FROM aliases WHERE alias = ? AND address != ? LIMIT 1',
+    [msg.payload.alias, message.address]
+  );
+  if (existing) {
     return Promise.reject('alias is already linked to another address');
   }
 
--- a/src/writer/alias.ts
+++ b/src/writer/alias.ts
@@ -14,10 +14,9 @@ export async function verify(message): Promise<any> {
     return Promise.reject('alias cannot be the same as the address');
   }
 
-  const [existing] = await db.queryAsync(
-    'SELECT address FROM aliases WHERE alias = ? LIMIT 1',
-    [msg.payload.alias]
-  );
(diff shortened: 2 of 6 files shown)


ChaituVR (changes requested):
I think we should add Unique index on DB instead of a check on application, wyt?

wa0x6e:
@copilot fix the merge conflict

wa0x6e:
@copilot fix the merge conflict

ChaituVR:
Looks good to me, except there are currently 120 alias addresses linked to multiple owners in the DB

```
SELECT alias, COUNT(DISTINCT address) as owner_count, 
  MIN(FROM_UNIXTIME(created)) as first_created, 
  MAX(FROM_UNIXTIME(created)) as last_created 
FROM aliases 
GROUP BY alias 
HAVING owner_count > 1 
ORDER BY owner_count DESC, last_created DESC
```

How do we handle this before running the `ALTER TABLE aliases ADD UNIQUE KEY alias (alias);` ?

wa0x6e:
We did not prevent multiple alias before this PR, and allows multiple addresses using the same alias.

Aliases generation on our UI should generate a new alias each time, and never reuse, so those duplicate probably come from people who set them via API, and use this a kind of delegation system, the duplicate alias was a feature. Changing the behavior to be unique will probably break their integration.

On the other hand, with the new UI page for creating alias, with the alias address only set via the URL, anybody can set an alias address they control in the url, and share this page. Once an user sign this page without paying attention (because this alias address is new to him), he hereby authorize this alias to act on his behalf.

Merging this PR may break integration for the existing integrator using it as a delegation system, but not merging it may expose the alias to phishing attacks.

The patch should probably be on the UI alias creation page, and make the page params more robust by adding a verification message signed the alias wallet, and be time sensitive (more robust way would be to also include the aliased address in the verification message)

ChaituVR:
Oh yes i agree. but what i meant was that if we run `ALTER TABLE aliases ADD UNIQUE KEY alias (alias);` it will fail no? because there's already duplicate data, what do we do with the data 🙈

wa0x6e:
Maybe not go with the PR, as duplication is widely used as a feature

ChaituVR:
But the phishing attack you mentioned is super valid 🙈 Maybe we can ignore uniqueness at SQL level and still check in app level? so weird that so many already use the alias system in wrong way. 🙈 not sure what's the right way to handle this. maybe we need need two signatures (one from user and one from alias as you suggest) which are signed at same time, to prove both belong to same user?

wa0x6e:
Duplicates have been removed from the database, they were not genuine. 

Database has also been updates with the new index

ChaituVR (approved):
tAck

PR #634, opened 2026-04-23
refactor: always allow alias for all spaces

bonustrack:
Closes snapshot-labs/workflow#793

## Summary
- Drop the `space.voting.aliased` opt-in — every space now accepts `vote`, `proposal`, `update-proposal`, `delete-proposal` signed by a valid alias of the author.
- Promote `flag-proposal` to EVM (it was already allowed on Starknet via `TYPES_EXECUTABLE_BY_STARKNET_ALIAS`).

Kept the diff intentionally minimal (4 lines); the `space.voting.aliased` field on existing space settings is simply ignored, no migration needed. The three-bucket structure in `alias.ts` and the `optionalAlias` param stay for now — a follow-up PR can collapse them.

## Test plan
- [x] `yarn lint`
- [ ] `yarn test` (needs DB — run in CI)
- [ ] Manual: with an `(address=A, alias=B)` row in `aliases`, submit a `vote` signed by B with `message.from = A` to a space **without** `voting.aliased`. Before: rejected. After: accepted.
- [ ] Manual: submit a `flag-proposal` signed by an EVM alias of a space admin. Before: rejected. After: accepted.
- [ ] Manual: submit a `vote` signed by B when `(A, B)` is missing from `aliases` → still rejected with `wrong alias` (regression check).

## Follow-ups (separate PRs)
- Collapse the three type buckets in `alias.ts` into one flat list; drop the now-always-`true` `optionalAlias` parameter.
- Add `signer` column to `votes` / `proposals`, populated with `body.address` when it differs from `message.from` (#793).
- Add `type` (`user` | `agent`) to the `aliases` table (#793).
- Remove `aliased` from the `voting` object in the snapshot.js space JSON schema.

### DIFF
--- a/src/helpers/alias.ts
+++ b/src/helpers/alias.ts
@@ -20,7 +20,8 @@ const OPTIONAL_TYPES_EXECUTABLE_BY_ALIAS = [
   'vote-string',
   'proposal',
   'update-proposal',
-  'delete-proposal'
+  'delete-proposal',
+  'flag-proposal'
 ] as const;
 
 // These types can be executed with a Starknet alias
--- a/src/helpers/alias.ts
+++ b/src/helpers/alias.ts
@@ -1,20 +1,14 @@
-import { uniq } from 'lodash';
 import db from './mysql';
 
 const DEFAULT_ALIAS_EXPIRY_DAYS = 30;
 
-// These types can always be executed with an alias
 const TYPES_EXECUTABLE_BY_ALIAS = [
   'follow',
   'unfollow',
   'subscribe',
   'unsubscribe',
   'profile',
-  'statement'
-] as const;
-
-// These types can be executed with an alias only when enabled in the space settings
-const OPTIONAL_TYPES_EXECUTABLE_BY_ALIAS = [
+  'statement',
   'vote',
   'vote-array',
   'vote-string',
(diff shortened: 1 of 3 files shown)


src/helpers/alias.ts
ChaituVR, line 1: We don't want to track the signer on `vote`, `proposal` etc? maybe in next PR?

test/unit/helpers/alias.test.ts
ChaituVR, line 1: we can add more test cases:


- verifyAlias('vote', ...) with a valid alias in DB → resolves.
- verifyAlias('proposal', ...) with a valid alias in DB → resolves.
- verifyAlias('update-proposal', ...) with a valid alias in DB → resolves.
- verifyAlias('delete-proposal', ...) with a valid alias in DB → resolves.
- verifyAlias('flag-proposal', ...) with a valid EVM alias in DB → resolves (covers the Starknet → EVM promotion).
- verifyAlias('vote', ...) when DB returns empty → rejects 'wrong alias'.
- verifyAlias('settings', ...) with an alias pair → rejects 'alias not allowed for the type: settings' (already covered, keep).
- verifyAlias('alias', ...) via alias signer → rejects (alias registration itself can't be aliased).
- verifyAlias('vote', ...) where body.address === message.from → resolves without touching DB (no queryAsync call).

ChaituVR (changes requested):

src/helpers/alias.ts
bonustrack: This would require change in db but I want us to move faster on this and get this merged sooner than later. So yes, that would fit into another PR.

ChaituVR (approved):
utAck

PR #636, opened 2026-04-28
fix: restrict delete-proposal alias to Starknet authors

bonustrack:
## Summary
- Drop `delete-proposal` from `TYPES_EXECUTABLE_BY_ALIAS`.
- Gate it behind a Starknet author check via a small `TYPES_EXECUTABLE_BY_STARKNET_ALIAS = ['delete-proposal']` list and reintroduce `isStarknetAddress`.
- EVM aliases can still vote / propose / update-proposal / flag-proposal — only deletion is restricted.

Follows up on #634, which inadvertently promoted `delete-proposal` to EVM aliases when collapsing the type buckets. Deletion is destructive and irrecoverable, so the pre-refactor Starknet-only behavior is the safer default until alias-typed signers (user vs. agent, per snapshot-labs/workflow#793) land.

## Test plan
- [x] `yarn lint`
- [x] Added unit case: `delete-proposal` from an EVM alias rejects with `alias not allowed for the type: delete-proposal`.
- [ ] `yarn test` (CI)
- [ ] Manual: EVM alias signing `delete-proposal` for an EVM author → rejected.
- [ ] Manual: Starknet alias signing `delete-proposal` for a Starknet author with a valid alias row → accepted.
- [ ] Manual: EVM alias signing `vote` / `proposal` / `update-proposal` / `flag-proposal` → still accepted (regression).

### DIFF
--- a/src/helpers/alias.ts
+++ b/src/helpers/alias.ts
@@ -2,7 +2,7 @@ import db from './mysql';
 
 const DEFAULT_ALIAS_EXPIRY_DAYS = 30;
 
-const TYPES_EXECUTABLE_BY_ALIAS = [
+const TYPES_EXECUTABLE_BY_ALIAS: readonly string[] = [
   'follow',
   'unfollow',
   'subscribe',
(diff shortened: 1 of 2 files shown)


ChaituVR:
@copilot+claude-opus-4.7 Add these tests:
- resolves for delete-proposal when the author is a Starknet address with a valid alias 
- rejects with 'wrong alias' when the type is allowed but no alias row exists

ChaituVR (approved):
utAck

wa0x6e (approved):
utAck

PR #638, opened 2026-05-04
chore: migrate package manager from Yarn to Bun

bonustrack:
## Summary
- Swap `yarn.lock` → `bun.lock`; `package.json` scripts now invoke `bun run` / `bunx` instead of `yarn`
- Pin `prettier`, `@snapshot-labs/eslint-config`, `@snapshot-labs/prettier-config` to exact versions and add a single `overrides` entry for `@snapshot-labs/eslint-config-base@0.1.0-beta.18` to preserve the resolutions yarn had (Bun's stricter prerelease semver would otherwise pull newer stables that break lint cleanliness)
- Point CI workflows at `snapshot-labs/actions@feat/bun-support` with `package_manager: bun`

## Notes
- `bun install`, `bun run lint`, and `bun run typecheck` pass cleanly locally
- CI on this branch will stay red until snapshot-labs/actions#`feat/bun-support` is fixed: that PR currently uses `bun test` / `bun build` in the reusable workflow, which invoke Bun's built-in test runner / bundler instead of the package.json scripts. They need to be `bun run test` / `bun run build`
- Once the actions PR lands on `main`, repoint the three workflow files from `@feat/bun-support` → `@main`
- Holding off on dropping `ts-node`/`nodemon` (Bun could replace them natively) for a later pass to keep this diff minimal

## Test plan
- [ ] `bun install` produces a clean `bun.lock`
- [ ] `bun run lint` passes
- [ ] `bun run typecheck` passes
- [ ] `bun run test:setup && bun run test` passes against local MySQL/Redis
- [ ] CI lint job green after actions PR merges
- [ ] CI test job green after actions PR merges
- [ ] CI sentry-release job green on master after actions PR merges


bonustrack:
This PR is deployed in production in livenet and testnet already

package.json
```diff
@@ -33,13 +33,14 @@
     "@snapshot-labs/snapshot-metrics": "^1.4.1",
     "@snapshot-labs/snapshot-sentry": "^1.5.5",
     "@snapshot-labs/snapshot.js": "^0.14.20",
+    "ajv": "8.11.0",
```
wa0x6e, line 37: Why this new dependency ?

wa0x6e: Ajv is not used directly by sequencer, and will be cleaned up by a agent in the future.

@tony8713 what is adding this dependency resolving ? is there other way to do it ?

wa0x6e (approved):
tAck

Not a fan of this undocumented ajv dependency hack, as it can be cleaned in a future PR by someone not aware of this workaround.

A cleaner way would be to just upgrade s.js to use the newer ajv version

tony8713:
@wa0x6e following up on the ajv override comment — opened https://github.com/snapshot-labs/snapshot.js/pull/1197 to pin `ajv` exact at `8.11.0` inside snapshot.js itself. Once that ships in 0.14.21 I'll bump snapshot.js here and drop the direct `ajv` dependency from this PR.

Quick context on why pinning rather than upgrading: I ran the bisect locally (bun + jest + ts-jest, like CI). `ajv-errors@3.0.0` emits `{"str":"errN"}.keyword` instead of `errN.keyword` in compiled validators when paired with ajv@8.20.0, throwing `SyntaxError: Unexpected token ':'`. ajv-errors has no newer release and there's no working ajv 8.21+ yet, so until one of those moves we can't actually use a "newer" ajv. Pinning at the snapshot.js source means downstream services don't each need to repeat the override.

tony8713:
Pushed 08c838b on this branch — bumps `@snapshot-labs/snapshot.js` 0.14.20 → 0.14.21.

About the `ajv: 8.11.0` direct dep — I tried to drop it as part of this bump and it didn't quite work. Honest write-up:

**What 0.14.21 ships**
[snapshot-labs/snapshot.js#1197](https://github.com/snapshot-labs/snapshot.js/pull/1197) tightens the `ajv` constraint inside snapshot.js from `^8.11.0` to exact `8.11.0`. So now snapshot.js can never resolve to the broken `ajv@8.20.0` regardless of which package manager pulls it.

**Why the sequencer's `ajv` dep still has to stay (verified locally)**
With the bump but the direct `ajv` dep removed:
- `node_modules/ajv` resolves to **6.15.0** (eslint's transitive copy lands at root because nothing else fights for that slot).
- snapshot.js, ajv-formats, ajv-errors each get their own nested `ajv@8.11.0`.
- They're the right *version* now but they're three different *constructor identities*, which breaks `ajv-errors`'s vocabulary registration — same `SyntaxError: Unexpected token ':'` from compiled validators. 63 unit tests fail.

With the bump + the direct `ajv: 8.11.0` retained:
- `node_modules/ajv` is `8.11.0` at root, shared by snapshot.js / ajv-formats / ajv-errors.
- `eslint/node_modules/ajv` and `@eslint/eslintrc/node_modules/ajv` stay nested at 6.15.0.
- Schema-validation tests (`spaceValidation.test.ts`, `envelope.test.ts`) pass; `lint` + `typecheck` clean.

I also tried `"overrides": { "ajv": "8.11.0" }` at the top level — it forces the override into eslint's tree too, eslint can't find a 6.x ajv, and `bun run lint` crashes. Scoped overrides like `"@snapshot-labs/snapshot.js > ajv"` aren't supported by bun yet ([oven-sh/bun#6608](https://github.com/oven-sh/bun/issues/6608)).

So the snapshot.js pin is now defense-in-depth (no consumer can accidentally pull `ajv@8.20.0`), and the sequencer's direct `ajv` dep is the deduplication lever — which is the part bun's hoisting still requires until #6608 ships. Happy to add a code comment near the dep referencing this, if useful.

wa0x6e (approved):
tAck

